Issue #13, opened 2017-08-08
Maven/Gradle?


ava1ar:
Hi, guys!

Do you use Maven or Gradle for managing the dependencies and build process? It is really strange to see modern Java application, which doesn't use one of them.

Shusshu:
I'm in the process of adding support for Maven but I saw code is missing from the repo...

They are including a bunch of plugins in the jar directly which I guess is not right

Shusshu:
This can be closed :)

PR #15, opened 2017-08-08
Feature/maven setup

Shusshu:
https://github.com/BiglySoftware/BiglyBT/issues/13

It's not exactly the same jar produced by the installer - the latter includes plugins that is not part of the github code base

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,3 +132,6 @@ core/lib/libWIN32Access/ipch/
 .idea/
 /uis/src/tux/
 *.DS_Store
+
+#maven
+/target
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -1,0 +1,86 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<artifactId>biglybt-core</artifactId>
+	<packaging>takari-jar</packaging>
+	<name>BiglyBT - Core</name>
+
+	<parent>
+		<groupId>com.biglybt</groupId>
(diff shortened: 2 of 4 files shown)


TuxPaper:
Thanks for starting this!

It would be good to get the two jars as close as possible. What exactly did you find included in the installer's biglybt.jar that's not part of the github codebase?

afaik, we just use https://github.com/BiglySoftware/BiglyBT/blob/master/core/BiglyBT_parg.jardesc to build the jar, then jarsign it.

Shusshu:
After fixing the maven build that's what I have now:

The following packages are missing from the code base:
package com.biglybt.core.neuronal
package com.biglybt.core.speedmanager.impl.v3


And these packages are part of the build with maven but not from the installer
package com.biglybt.core.util.spi;
package com.biglybt.core.util.dns;
package com.biglybt.core.util.jman;
package com.biglybt.core.xml.util;
and many others but most of them are just empty classes

TuxPaper:
> The following packages are missing from the code base:
> package com.biglybt.core.neuronal
> package com.biglybt.core.speedmanager.impl.v3

These have been removed since the 1.0.0.0, so everything is good there.

> And these packages are part of the build with maven but not from the installer
> package com.biglybt.core.util.spi;
> package com.biglybt.core.util.dns;
> package com.biglybt.core.util.jman;
> package com.biglybt.core.xml.util;

hmm, I'm seeing files in the release jar
image

I do know some/most of those packages you listed are invoked by reflection, but I don't think that makes a difference.

Shusshu:
> These have been removed since the 1.0.0.0, so everything is good there.

Well it's still in the code base on github ;)

For the rest I think it's fine.

Maybe you can see for yourself the jar diff by running 
`mvn clean package`

TuxPaper:
Definitely not on Github! https://github.com/BiglySoftware/BiglyBT/commit/2a6641f78e3dff2bae6f701983385da8addce5ee

But that's really unimportant to this, so I'll approve the pull.  Hopefully it's as easy as clicking that lovely green button, although the drop down arrow on it is suspicious ;)

PR #18, opened 2017-08-09
Fix Jar manifest classpath + cleanup

Shusshu:


TuxPaper:
Thanks for the commits! :)

Issue #19, opened 2017-08-09
Plugins


Shusshu:
I'd like to migrate my plugins to BiglyBT.

Could you provide the API jar on the github repo maybe as a side module that can act like a maven repository ? or publish to maven central ? or worse case a jar and I'll do as before ;) 

https://github.com/Shusshu/azureus-plug-globalinfo
https://github.com/Shusshu/azureus-plug-timedautostart 
Note: This is extremely ugly code from 10-12 years ago 

I'll try to figure a way to produce the API jar via maven and the current code base.... (assuming I only export the package ...plugin...  only interfaces) (Probably with http://maven.apache.org/plugins/maven-assembly-plugin/plugin-info.html )

TuxPaper:
@parg  does the API jars, so I'll let him handle that.

But I do have a TimedAutoStart with changes, since the source was under http://svn.vuze.com/public/plugins/TimedAutoStart as well.  I've attached a diff, and can probably do a pull request once I figure out how that actually works in IntelliJ ;)

[TimedAutoStart.diff.txt](https://github.com/BiglySoftware/BiglyBT/files/1212357/TimedAutoStart.diff.txt)

We never had the GlobalInfo plugin on http://svn.vuze.com/public/plugins/ so unfortunately, it didn't get into the refactoring tool or manual migrations I made.  (fwiw, all the plugins in http://svn.vuze.com/public/plugins/ are ready to be biglybt plugins, we just chose to limit the biglybt plugin list to ones we knew people used, since there's a lot of dead ones in that list)

Shusshu:
I don't mind updating it myself, well I won't cleanup the dirty ugly code but at least make it work on BiglyBT.

I couldn't figure a way to do the plugin-api.jar with maven

TuxPaper:
I imagine we'll be using something similar to what's in https://svn.vuze.com/public/client/trunk/build.xml

Most importantly,
`
	<patternset id="pluginapi.inc">
		<include name="org/gudy/azureus2/plugins/**" />
		<include name="org/gudy/azureus2/ui/swt/plugins/**" />
		<include name="org/gudy/azureus2/ui/swt/views/*IView*" />
		<include name="org/gudy/azureus2/ui/swt/IconBarEnabler*" />
	</patternset>	
`

Which is all now `com/biglybt/pif` and `com/biglybt/ui/swt`.. although I don't think we need the last two anymore.

And here's the relevant jar bundling snippet
`
		
[...]

PR #22, opened 2017-08-10
Plugin Api generation via ant inside maven

Shusshu:
resolves https://github.com/BiglySoftware/BiglyBT/issues/19

Finally I couldn't manage to do it via maven only so it's via ant inside maven :/

I'm sure someone can improve it to get rid of the ant part :)  (At least we don't need to install ant)

### DIFF
--- a/uis/pom.xml
+++ b/uis/pom.xml
@@ -127,4 +127,49 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>plugin-api</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-antrun-plugin</artifactId>
+                        <version>1.8</version>
+                        <executions>
+                            <execution>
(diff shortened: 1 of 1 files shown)


Shusshu:
Oh wait before merging this... it seems some classes are missing

Shusshu:
This was missing
<include name="com/biglybt/ui/common/UIInstanceBase.*" />

(I also updated my 2 plugins: https://github.com/Shusshu/biglybt-plugin-globalinfo & https://github.com/Shusshu/biglybt-plugin-timedautostart ) if you want to put them somewhere else for distribution ...

TuxPaper:
Boo on `UIInstanceBase` being needed for the plugin interface. There was no need for `UIInstanceBase` to be on `UISWTInstance` when it could have been on `UISWTInstanceImpl`.  I'll remove it for the next releasing, assuming the change doesn't require rebuilding every plugin.

PR #56, opened 2017-09-14
Prepare next development iteration

Shusshu:
Do you want to try the maven plugin to do releases via cmd line?

http://maven.apache.org/maven-release/maven-release-plugin/

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -8,7 +8,7 @@
 	<parent>
 		<groupId>com.biglybt</groupId>
 		<artifactId>biglybt-parent</artifactId>
-		<version>1.0.0</version>
+		<version>1.0.2.1-SNAPSHOT</version>
 	</parent>
 
 	<dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>com.biglybt</groupId>
 	<artifactId>biglybt-parent</artifactId>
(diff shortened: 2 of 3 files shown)


parg:
Not sure I'm at the point where I need to look into release automation yet - possibly something for the future though as things grow and it becomes more than a simple task...

Shusshu:
I can look into linking the version in Constants.java with maven maybe....
And this does more stuff. It creates a release with a git tag and then make a new version for the next development iteration... it's pretty cool and avoid mistakes

But no problem you can also refuse this one as this is not mandatory

TuxPaper:
No issues with this pull, since it's just a manual update of the version number.  The automating discussion is a different matter :)

parg:
We used to have a Jenkins/Hudson task that generated the 'beta' builds we produce between releases (the 1.0.0.1_Bnn ones). This modified the

`public static final String AZUREUS_VERSION  = "1.0.2.1_B01";`

in Constants.java along with

`public static final String BUILD_VERSION  = "@build.version@";   //Ant replace `

which was quite handy (nothing was committed/tagged though)

It managed build failures etc so that the successful builds were correctly named B01, B02,...

Shusshu:
We have 3 possibilities:
1. The same as Ant: https://stackoverflow.com/a/8817516/1228221
2. Reading a properties https://stackoverflow.com/a/2470071/1228221
3. What I usually do in my maven projects: (template filtering): https://stackoverflow.com/a/7161303/1228221

PR #57, opened 2017-09-14
Fix Javadoc

Shusshu:
Some javadoc issues (there are more but well :))


TuxPaper:
haha, thanks.  Yeah, there's probably a ton of invalid parameter names, because (older?) Eclipse never warned about them after changing a parameter name.

PR #58, opened 2017-09-14
BiglyBT @since API reset to 1.0.0.0

Shusshu:
Useful for plugin 

I don't know if you want to keep the previous version but it seems odd to keep it...

I used a regex to find & replace: @ since (\d)\.(\d)\.(\d)\.(\d)

edit: @since sorry for the mention

### DIFF
--- a/core/src/com/biglybt/activities/ActivitiesEntry.java
+++ b/core/src/com/biglybt/activities/ActivitiesEntry.java
@@ -551,7 +551,7 @@ public class ActivitiesEntry
 	 *
 	 * @return Only returns TOTorrent set via {@link #setTorrent(TOTorrent)}
 	 *
-	 * @since 3.0.5.3
+	 * @since 1.0.0.0
 	 */
 	public TOTorrent getTorrent() {
 		return torrent;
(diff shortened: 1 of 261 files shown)


TuxPaper:
hmm, I'll have to think about this one.  On the one hand, I understand your reasoning.  On the other hand, it's a whole lot of lost `@since` version history.  But do the old `@since` have any use to anyone? probably not.

My first thought was changing it from `@since 2.5.0.4` to `@since Azureus 2.5.0.4`, but we don't want to use that word, so it would have to be `@since AZ 2.5.0.4`.  Which still begs the question, is it useful to have a history of Azureus `@since` versions?  Or the even deeper question : Do we need `@since` at all, aside from in the plugin interface?

Shusshu:
Yes you would lose a bit of the history but this is also a new beginning - Sometimes you have to cut ties completely. You decide I don't mind if you refuse the PR (it's less than 5 minute of work)

I remember when I first made my plugin many years ago it was useful to know which API I should avoid to get as much compatibility as possible

TuxPaper:
I think I'd rather remove all the @since from the non-plugin interface classes, and have the plugin interfaces all at 1.0.0.0.
But it's pretty low priority.

PR #59, opened 2017-09-14
Add Travis CI

Shusshu:
See example here https://travis-ci.org/Shusshu/BiglyBT/builds/275623617

This will run on every PR to ensure the build compiles and run tests (We might need to do some setup or reorganise the test location to automatically run them)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,4 @@
+language: java
+
+jdk:
+  - oraclejdk8


TuxPaper:
Lordy, do we have tests!?!?!? ;)

I'll merge this when I have a moment to link travis-ci.org to our github account.

In the past I've just relied on parg complaining the source doesn't compile after I commit something, so this might be a step up ;)

Shusshu:
I'm sure it's still a good _local_ system ;)

davisjackie321:
Yo

davisjackie321:
Nice

davisjackie321:
Go

davisjackie321:
Nice

yusdiana:
H

ferdnyc:
Hey, so I couldn't help but notice that @Shusshu 's Travis CI instance for BiglyBT [is still running builds _anyway_](https://travis-ci.org/Shusshu/BiglyBT/builds). But they're all failing, since without a `.travis.yml` config file to tell it otherwise, it defaults to trying to build the project as Ruby source.

Since having that file in the tree has literally no effect except when Travis CI is running builds, there can't be any harm in merging this, right? Even if Travis isn't set up in GitHub, it just means the source will be ready for that to happen.

parg:
Did that fix anything?

Shusshu:
Yes but it's not building ^^  Gotta figure that out now

Shusshu:
Yes that worked out :)

PR #257, opened 2018-01-07
translating to Slovenian

lkure:
added new translations and corrected many existing ones


BiglyBT:
Thanks for the translations!  The Crowdin translation service will probably (annoyingly) re-order the keys, but I'll make sure it doesn't remove your changes or new entries.

Issue #545, opened 2018-07-21
"Search for Existing Data Files" progress window is twitchy


ferdnyc:
Oh yes, one other minor-ish display thing:

Whenever the "Search for Existing Data Files" process runs, after setting up the paths and options in the initial dialog and pressing OK, it opens up... I guess a log window, to show the progress of the search while it's scanning files.

At least on my system with the way I have files organized, the pathnames of the files being processed are invariably too long for those log messages to fit inside the window. As a result, it spends the entire scanning run twitching madly back and forth, jerking all the way right to the end of the line, then jerking back left to the start of a line, and repeat at a rapid pace until the last file is scanned.

I don't really have a screenshot since you'd have to see it in motion, but hopefully you've seen what I'm referring to.

Not only is the constant jerking left and right disconcerting, it makes the actual output impossible to read until the scan is complete.

If the messages can't be wrapped or reformatted to fit inside the window, then I think just _not_ autoscrolling left and right would be a perfectly reasonable solution. Just let the long lines extend past the right edge of the window, and the user can scroll over if they need to see the rest of the text.

#### System details:
```
Java 1.8.0_172 (64 bit)
  Oracle Corporation
/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.172-9.b11.fc28.x86_64/jre

SWT v4757, gtk/3.22.30
Linux v4.17.3-200.fc28.x86_64, amd64 (64 bit)
B1.5.0.1_B23/4 az3
```

ferdnyc:
I saw that the _B24 release notes mentioned fixes for this issue, but unfortunately it doesn't really seem to be changed much, at least on Linux.

I have a screencast of one session, excuse the crazy colors meant to obscure my file names/paths:

![bigly](https://user-images.githubusercontent.com/538020/43092303-bf17ee06-8e7a-11e8-8cc1-c6bc747951f8.gif)

It seems to play OK here as an animated GIF, but just in case here's a zipfile containing the MP4 it was converted from.

[Bigly.zip](https://github.com/BiglySoftware/BiglyBT/files/2220372/Bigly.zip)

ferdnyc:
The thing I find really weird, looking at that GIF, is that even _after_ the lines are short enough to fit in the window, it **still* twitches back and forth a bit... though not as far. It seems as if the longest output line sets the "resting" scroll position, and it keeps trying to get back _there_. But showing each subsequent line forces it back to the start.

[...]

PR #641, opened 2018-10-10
Add wrapping to TextViewerWindow (for "...Existing Files")

ferdnyc:
So, I found a way to solve the `Search For Existing Files...` twitching. It's kind of brute-force, but it works.

I just turned wrapping on for the `TextViewerWindow`. 

(Actually, I turned wrapping on for _all_ `TextViewerWindow` instances, since that option wasn't overrideable from the caller. So, I'm not sure what the side effects of that would be; depends where else `TextViewerWindow` is used.)

Like I said, brute-force. Your call on whether it's the _right_ fix. (And I'm fully anticipating that it isn't, so don't hesitate to reject this PR. But I figured maybe it'll trigger some inspiration for a _right_-er fix.) I wanted to report that it _does_ solve the twitching issue, anyway.

Fixes: #545

### DIFF
--- a/uis/src/com/biglybt/ui/swt/TextViewerWindow.java
+++ b/uis/src/com/biglybt/ui/swt/TextViewerWindow.java
@@ -129,7 +129,7 @@ public class TextViewerWindow {
     gridData.horizontalSpan = 3;
     Utils.setLayoutData(label, gridData);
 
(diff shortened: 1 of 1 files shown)


parg:
If it works for you I'll give it a go - if 'others' complain then we can take another look

ferdnyc:
Or I can fight them! :boxing_glove: :grinning:

PR #661, opened 2018-10-19
Update issue_template.md

KonoromiHimaries:


parg:
Hi,
What's the purpose of this proposed change?
Thanks
Parg

TuxPaper:
Looks like it hides our template text (when not in a textbox) if the user forgets to remove it.  Can't see any harm in trying it.

PR #761, opened 2018-12-19
Add 'SHA-1' to 'hash', where it's needed

Tobias-B-Besemer:
'SHA-1' should be added to term 'hash', when it's a 'SHA-1 hash', so the user know what kind of hash it is.
Expecting that the user reads in Wikipedia about e.g. Magnet Links, is not really 'nice'.
So: Add 'SHA-1' to 'hash', where it's needed!

### DIFF
--- a/core/src/com/biglybt/internat/MessagesBundle.properties
+++ b/core/src/com/biglybt/internat/MessagesBundle.properties
@@ -3395,7 +3395,7 @@ Button.open=Open
 
 plugin.init.load.failed=Error loading plugin '%1' from %2
 plugin.init.load.failed.classmissing=Uninstall the plugin via 'Tools->Plugins->Uninstallation Wizard' and then re-install it.
-v3.MainWindow.search.tooltip=Enter your search terms, magnet link, hash or torrent URL and hit enter!
+v3.MainWindow.search.tooltip=Enter your search terms, magnet link, SHA-1 hash or torrent URL and hit enter!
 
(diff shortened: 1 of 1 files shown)


parg:
Thanks for the suggestion but we're going to stick with just 'hash' for the moment.

Tobias-B-Besemer:
But a MD5 hash doesn't work in this box, right?

parg:
If you know enough about things to consider entering a hash then you should know that torrent files use SHA1 in my opinion, or you can figure it out (as you have)

TuxPaper:
Saying 'hash' is future proofing it, imo, and infers "torrent hash" which is currently SHA-1, but someday may include other hashes.  It's unclear if the hash for merkle-tree torrents is a SHA-1 hash anymore (I only scanned BEP-30 for it, and there was no clear "info hash is still SHA-1" in the spec), so if we ever support that BEP, we'd also support whatever hash format it uses.

FWIW, we accept hex SHA-1 and base32 SHA-1 hash.

I appreciate your pull request, and your reasoning for adding the distinction.  Please don't be discouraged by our replies.. your pull request is a totally valid and logical one!

Part of my resistance is also the fact that any changes in the english language file causes all other languages to have that entry entirely wiped out.  Translators would have to translate the two lines again, and for the multiple languages that have no active translator, that will never happen. (also fwiw, this is a limitation/feature with the localization service we use, which annoys me to no end)

PR #901, opened 2019-03-03
I10n single line properties

lislei:
Single line property values in i10n property files / resource bundles. This harmonization accommodates easy diff between root and your localization of choice.
It seems like crowdin converts trasnlated values to single line values as default.


TuxPaper:
Thanks for doing all this work.  I noticed the extra line bug in Crowdin, but didn't get very far in detailing it out.  I'll try to merge it and import it into Crowdin, but it may take a few tries, as Crowdin seems to hate translations not coming from their editor.

lislei:
You're welcome. I didn't commit changes to the arabic and other languages as I lack experience with bidirectional text.
 
If anyone would advice me on how to restructure the semantics I would give it a second try!

TuxPaper:
On of the translation files we have the biggest problems with is the RCM one 
https://github.com/BiglySoftware/BiglyBT-plugin-aercm/tree/master/com/aelitis/plugins/rcmplugin/internat

There's some mega-long entries in there that are split with newlines, and everytime we do something outside of Crowdin, it doubles the newlines and I have to go in and muck with CrowdIn until I get it back to normal.

It sounds like you have a good system for doing these changes, I'd really appreciate help with that one.

lislei:
After having playing arond with [javas html](https://docs.oracle.com/javase/tutorial/uiswing/examples/components/index.html#HtmlDemo) demo i came up with this:
cyrillic_java_html

It seems like java is choosing text direction from intepreating the characters. Latin chars gives a different direction than cyrillic ones. When typing and moving the text cursor it switches symbol going from one text direction to the other. 

Pasting the same text into google translate seems to give you similar results. 

I'll give the Cyrillic translations another try.

lislei:
There seems to be an issue when mixing Latin and Cyrillic words (and symbols) when it comes to text direction. I have no way of verifying any changes I make will produce the desired outcome. 
One way to handle it would be to implements some tests that reads from the resource bundle and verifies the transfomed text as an utf-8 sequence.

lislei:
After some further investigation I concluded it to be a display issue. The method of working with property files and bi-directional text seems to be correct.
I found that the order of chars are the same you read and write them. This holds for RTL languages as well. Only difference is in the way the text is displayed.

I completed the work - see #926

PR #936, opened 2019-03-17
New Norwegian translations

lislei:
I gathered some of my work in this push. 
Enjoy!


core/src/com/biglybt/internat/MessagesBundle_no_NO.properties
```diff
@@ -2117,15 +2139,17 @@ PiecesView.DistributionView.weHave=Bruddstykker du har
...
+unix.script.new.title=Nytt oppstartsskript til {base.product.name} er tilgjengelig
+unix.script.new.text=Nytt {base.product.name} oppstartsskript er tilgjengelig og lagret til '%1'. \n\nDet er h\u00F8yst anbefalt at du avslutter {base.product.name} n\u00E5 og bytter til dette skriptet ('%2'). \n\n\Dersom du har gjort store endringer i ditt {base.product.name} skript se <A HREF="{unix.script.new.manual.url}">{base.product.name} Wiki: Unix Script</A>.\n\nDersom du installerte {base.product.name} fra en distro (yum, apt-get, ol.) anbefales det \u00E5 re-installere {base.product.name} via pakken fra <A HREF="http://www.biglybt.com/download/">{base.product.name} nedlastinger</A> (Du vil fortsatt ha det samme UI som du har n\u00E5)
```
lislei: The download link should be normalised here. 

Would the right approach be to define a constant in `com.biglybt.core.util.Constants` 
and wire it in by `MessageText.DEFAULT_EXPANSIONS`?

parg: Sounds like a good approach (perhaps make it https as well). The annoying problem of all translations being invalidated if the main bundle is updated will raise its head if you change that though :(

lislei: Please elaborate.

lislei: @TuxPaper described to me an issue with the CrowdIn integration.
If updated the translated strings will be blank/lost.

From reading the crowdin documentations there is a configuration switch tor this case. 
[See `update_option` and `update_without_changes`](https://support.crowdin.com/configuration-file/#changed-strings-update).

TuxPaper: hmm, I wonder if this update_option is also available for their crowdin.yml config (Ours is https://github.com/BiglySoftware/BiglyBT/blob/master/crowdin.yml )

The doc you point to is their CLI, which is essentially their manual way to upload translations.  So what would happen is:
1. We'd commit your changes
1. Crowdin would sync, and erase all the non-english values for `unix.script.new.title`
1. We then have to manually re-upload all 30+ translations to Crowdin with the `update_without_changes` setting

Step 3 is pretty much what I do in their UI when I have need it to read all the github translations again.  It takes several hours on their servers to complete.  
There's also been issues in the past with translated, but unapproved strings in Crowdin, and the manual upload feature.  So, before Step 3, I go into every language in Crowdin and ensure all translations are approved, which takes anywhere from 5 min to 1/2 hour.

So, bottom line is, I can do a mass for all languages, it's just time consuming and error prone (ie. What happens if someone translates something in Crowdin while it's still updating manually?)

I'll do some testing with crowdin.yml today tho, and see what happens I add `update_option: update_without_changes`

lislei: I found the right documentation for the VCS: https://support.crowdin.com/vcs-integrations-configuring-online/ 
The same options are mentioned there :)

PR #940, opened 2019-03-19
Code Cleanup & optimized JDK8 syntax

lislei:
Removed some legacy code. See commit messages.

Is there any restrictions on using the full JDK 8 API?


TuxPaper:
The J8 restrictions are limited to Android, which we currently are still supporting API 15 (4.0.3 - Ice Cream Sandwich). Anything listed [here](https://developer.android.com/studio/write/java8-support#supported_features) that says "API 24" means we don't support it, including try-with-resources.  But that's not the full list :(  For example, StandardCharsets isn't available until API 19 according to https://developer.android.com/reference/java/nio/charset/StandardCharsets

With that in mind, unless you know of a difference between, `StandardCharsets.UTF_8` and `Charset.forName("UTF-8")`, I'd prefer the latter.  Similarly with  StandardCharsets.ISO_8859_1

If there's a valid reason for using an unsupported J8 feature, I can adjust the Android branch to use the older style, but I'd like to keep them in sync as much as possible.

TuxPaper:
Forgot to add, although Android doesn't support try-with-resources, we use them in our source.  You can see most of the difference between trunk and the Android branch are rewrites of try-with-resources, and the other 1/2 are pretty much Long.compare.  [Diff if you are curious](https://github.com/BiglySoftware/BiglyBT/compare/android)

TuxPaper:
Since you seem to be on a cleanup kick (and I love it), I should point out that I have a huge restructure of the ConfigSection pages, which also affects core and plugin config parameter classes.  So, don't be tempted to clean anything up there :)

lislei:
I'm thrilled over your feedback. I've added som documentation to `readme.md`.

I did a rewrite eliminating the use of StandardCharsets.

lislei:
Is there any changes in this PR that conflicts with your rewrite? 

I'd definitly consult prior to do any big changes, I'm thinking a small PR is a good starting point for further contributions.

TuxPaper:
No conflicts yet :)

In the future, it's better to have separate smaller pull requests, so that ones with no issues can be merged quickly instead of getting stalled by issues with unrelated code changes.

Looks like you'll run into another messagebundle/Crowdin problem with the latest changes. Changing `network.ipv6.enable.support` in the main bundle will cause Crowdin to erase all translations for that key, even though you fixed all the translations too.  Crowdin don't care.  
I don't know what the right solution would be.  There's probably a good chance that even making a new key, like `network.ipv6.enable.support.2` and filling in the other translations, would not end up in Crowdin.. but it might be worth a try. 
The only real solution I can think of is to learn CrowdIn's API, and code up something that would take a key, and force publish it to Crowdin for each language using the github files as the source. I've thought about it a few times, but each time it seemed more hassle than it's worth and I either opt to force the translators to translate again, or say "meh, the text is close enough as is"

TuxPaper:
Some other general comments not related to this pull, but good FYI (I should write a wiki page or something :P) :

Things like lambdas, generics, for-each (instead of `for (int i = 0;...`) are perfectly fine, but the general rule of "keep blame history clean as possible" applies. For example, if you are modifying a for loop anyway, changing it to for-each is encouraged.  But changing all for loops everywhere into for-each isn't something I'd want to see.  
Same applies to making Lists generic.. if you in the code messing with it, it's encourage to add a type to the List (assuming it won't break the Plugin Interface for existing plugins)
Similarly, lambdas, especially when we call `Utils.execSWTThread(new AERunnable() { ... })` can be changed to `Utils.execSWTThread(() -> { .. })` if you are already modifying the code.

lislei:
Parg mentioned some issues with touching up the translations [on another PR](https://github.com/BiglySoftware/BiglyBT/pull/936#discussion_r267488908).  I pointed out a possible configuration switch that might help.

I dont have any experience with Crowdin. Since there is a `i10n_master` branch and a `crowdin.yml` config file im guessing it uses the github integration. 
How does the update/synch process work?

lislei:
I share the "keep blame history clean as possible" mentality and keep your footprint of changed lines to an minimum approach. 
Worth mentioning is the option to right click on any annotated file in IntelliJ and select "Annotate previous Revision". This gives you a quick and easy way of showing what the implementation looked like prior to a particular change.

TuxPaper:
Saw your comment in #936 and commented there first, but I'll CC from there for easier access.  New notes after the horizontal line

hmm, I wonder if this update_option is also available for their crowdin.yml config (Ours is https://github.com/BiglySoftware/BiglyBT/blob/master/crowdin.yml )

The doc you point to is their CLI, which is essentially their manual way to upload translations.  So what would happen is:
1. We'd commit your changes
1. Crowdin would sync, and erase all the non-english values for `unix.script.new.title`
1. We then have to manually re-upload all 30+ translations to Crowdin with the `update_without_changes` setting

Step 3 is pretty much what I do in their UI when I have need it to read all the github translations again.  It takes several hours on their servers to complete.  
There's also been issues in the past with translated, but unapproved strings in Crowdin, and the manual upload feature.  So, before Step 3, I go into every language in Crowdin and ensure all translations are approved, which takes anywhere from 5 min to 1/2 hour.

So, bottom line is, I can do a mass for all languages, it's just time consuming and error prone (ie. What happens if someone translates something in Crowdin while it's still updating manually?)

I'll do some testing with crowdin.yml today tho, and see what happens I add `update_option: update_without_changes`

---

We use the GitHub Integration feature https://support.crowdin.com/github-integration/
That's the feature that's causing the lines be be blanked.  It automatically updates Crowdin when it detects changes in our GitHub translation files.  Well, maybe just the one file, and it ignores the other language files.

However, the GitHub Integration also uses a yaml config file (the crowdin.yml) which is why I'm going to try to add that `update_option` option to it and see what it does :)

lislei:
I see you have done some testing.  
Do you have any findings yet?

BiglyBT:
Unfortunately, it does nothing.  I put `update_option: update_without_changes` into the [config](https://github.com/BiglySoftware/BiglyBT/blob/master/crowdin.yml), changed an existing value in Github, and Crowdin creates a new pull request, reverting my change.

TuxPaper:
Unfortunately, it does nothing. I put update_option: update_without_changes into the config, changed an existing value in Github, and Crowdin creates a new pull request, reverting my change.

TuxPaper:
I have one more test I want to try:  Creating a new key in the main bundle, and in a translation, and see if Crowdin pulls in the translation.  Maybe it's smarter for newly added keys.

lislei:
How does it work if u make a new key together with some of the translations filled in.
I'm thinking of the refactor case where you construct a new key and refactor the text some how...

PR #959, opened 2019-03-24
Removed IPChecker static text from message bundle

lislei:
Inlined static IPChecker urls and service names.
+ Made l10n keys fully qualified for easy search.

Im sure Chrowdin will remove any superfluous keys for translations.


lislei:
I was planning on moving it into `MessageText.DEFAULT_EXPANSIONS` but the different implementations are for spesific IPChecker services and thereby static.

PR #960, opened 2019-03-24
l10n constants

lislei:
I tried the exercise of removing the constant value "BiglyBT" from the message bundles.
This is a system wide constant anyhow and shouldn't be translated into something different.

What I discovered was that this value could only be used for expansion and not on itself. 
This improvement introduces some default keys wich are expandable (like the others).

I am treating the wiki URLs are the same for all languages, at least that's how it currently is.  
If it were to be updated, then one had to call the method `updateProductName()`

Unit-tests are added for this change.


lislei:
I'm hoping to eliminate all url constants from the message bundle in a later PR.

PR #963, opened 2019-03-24
Extracted wiki url as constants

lislei:
Defined l10n constants for wiki urls in `MessageText`.
Created a `Wiki` enum class holding all the configured urls to wiki pages.

I'd like feedback on `Wiki` class, both placement and if structure is suitable.

I also added a test that produces this output:
```
[INFO] Running com.biglybt.core.WikiTest
asserting url exits https://wiki.biglybt.com/w/UG_Options#Advanced_Network_Settings ... OK**
WARNING: no wiki on topic Advanced_Network_Settings
WARNING: no anchor found #Advanced_Network_Settings
asserting url exits https://wiki.biglybt.com/w/AllTrackersView ... OK
WARNING: no wiki on topic DNS
asserting url exits https://wiki.biglybt.com/w/UG_Options#DNS ... OK**
WARNING: no anchor found #DNS
asserting url exits https://wiki.biglybt.com/w/Decentralized_Chat ... OK
WARNING: no wiki on topic Devices_iTunes_Tips
asserting url exits https://wiki.biglybt.com/w/Devices_iTunes_Tips ... OK*
WARNING: no wiki on topic FAQ_Legal
asserting url exits https://wiki.biglybt.com/w/FAQ_Legal ... OK*
WARNING: no wiki on topic Failed_Update
asserting url exits https://wiki.biglybt.com/w/Failed_Update ... OK*
asserting url exits https://wiki.biglybt.com/w/FAT32_file_size_limit ... OK*
WARNING: no wiki on topic FAT32_file_size_limit
asserting url exits https://wiki.biglybt.com/w/Friends ... OK
asserting url exits https://wiki.biglybt.com/w/Friends#public-profile ... OK
WARNING: no wiki on topic Interface_Format_Overrides
asserting url exits https://wiki.biglybt.com/w/UG_Options_Interface_Format_Overrides ... OK*
WARNING: no wiki on topic Pairing
WARNING: no anchor found #Pairing
asserting url exits https://wiki.biglybt.com/w/UG_Options#Pairing ... OK**
WARNING: no wiki on topic Privacy_View
asserting url exits https://wiki.biglybt.com/w/Privacy_View ... OK*
WARNING: no wiki on topic Restarting_Issues
asserting url exits https://wiki.biglybt.com/w/Restarting_Issues ... OK*
asserting url exits https://wiki.biglybt.com/w/Secure_Remote_Password ... OK*
WARNING: no wiki on topic Secure_Remote_Password
asserting url exits https://wiki.biglybt.com/w/Speed_Limit_Scheduler ... OK
WARNING: no wiki on topic SWT_Cant_Auto_Update
asserting url exits https://wiki.biglybt.com/w/SWT_Cant_Auto_Update ... OK*
asserting url exits https://wiki.biglybt.com/w/Unix_Startup_Script ... OK*
WARNING: no wiki on topic Unix_Startup_Script
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 19.875 s - in com.biglybt.core.WikiTest
```


core/src/com/biglybt/plugin/net/buddy/BuddyPlugin.java
```diff
@@ -1,2242 +1,2243 @@
...
-		}
-
-		protected String
-		getLastModified()
-		{
-			return( last_modified );
-		}
-	}
-}
+/*
```
lislei: My intellij shows only one added line and one changed line. 
I don't see whats wrong. Please advice!

lislei: Here's my diff 
bilde

TuxPaper:
Any reason the URLs are just String Constants?  

Seems like a lot of Object creation for what will only be used to Launch an external browser with a String.  Plus we always get complaints about startup time, and I think Strings are fairly fast compared to a set of new objects.

core/src/com/biglybt/plugin/net/buddy/BuddyPlugin.java
lislei: Seems like that file has CRLF style of line endings. 
This is what i see opening the raw file at https://raw.githubusercontent.com/BiglySoftware/BiglyBT/master/core/src/com/biglybt/plugin/net/buddy/BuddyPlugin.java

bilde

lislei:
Yes, you're right. That would be a simpler implementation.
I was unsure if there possibly can be more than one wiki or if one wiki can have several languages... 

Anyhow after reasessing it the proposed implementation wouldn't solve the case of multiple languages.
I'll simplify it as pointed out :)

core/src/com/biglybt/plugin/net/buddy/BuddyPlugin.java
lislei: I created a PR to fix the line endings in this repo. See #965

lislei: I re-added CRLF to spesific file as a workaround for this PR.

lislei: @parg - I suggest you set the autocrlf flag, Heres the command:
```
git config --global core.autocrlf true
```

uis/src/com/biglybt/plugin/startstoprules/defaultplugin/ui/swt/ConfigSectionDownloading.java
```diff
@@ -95,12 +95,11 @@ public Composite configSectionCreate(Composite parent) {
 	linkLabel.addMouseListener(new MouseAdapter() {
```
lislei: Isn't this the same as default behavour in LinkLabel?

lislei: I see the default behaviour is slightly different - it handles Strings
```
	    	new MouseAdapter() {
		      @Override
		      public void mouseDoubleClick(MouseEvent arg0) {
		    	  launch( arg0 );
		      }
		      @Override
		      public void mouseUp(MouseEvent arg0) {
		    	  launch( arg0 );
		      }
		      private void
		      launch(
		    	  MouseEvent	arg0 )
		      {
		    	  String url = (String) ((Label) arg0.widget).getData();
		    	  
		    	  if ( url != null ){
		    		  Utils.launch( url );
		    	  }
		    	  
		      }
		    }
```

Would be nice if the default implementation would be able to deal with URL's or Wiki enums somehow

uis/src/com/biglybt/plugin/startstoprules/defaultplugin/ui/swt/ConfigSectionDownloading.java
```diff
@@ -95,12 +95,11 @@ public Composite configSectionCreate(Composite parent) {
...
+			Utils.launch(((Wiki) arg0.widget.getData()).asURL());
 		}
-
 		@Override
 		public void mouseDown(MouseEvent arg0) {
-			Utils.launch((String) ((Label) arg0.widget).getData());
+			Utils.launch(((Wiki) arg0.widget.getData()).asURL());
 		}
 	});
 	ClipboardCopy.addCopyToClipMenu( linkLabel );
```
lislei: This looks quite similar to the default behavour in LinkLabel
```
	    ClipboardCopy.addCopyToClipMenu( label );

```

uis/src/com/biglybt/ui/swt/devices/DevicesFTUX.java
```diff
@@ -143,12 +143,12 @@ void open(Runnable fire_on_install) {
 		}
 
 		Link lblLearnMore = new Link(shell, SWT.NONE);
```
lislei: Can this be substituted with LinkLabel ?

uis/src/com/biglybt/ui/swt/mainwindow/MainStatusBar.java
```diff
@@ -613,7 +613,7 @@ public void handleEvent(Core core, Event e) {
 						ConfigSection.SECTION_CONNECTION);
 
 				if (PluginInitializer.getDefaultInterface().getConnectionManager().getNATStatus() != ConnectionManager.NAT_OK) {
-					Utils.launch(Constants.AZUREUS_WIKI + "NAT_problem");
```
lislei: There are many wiki pages that exists on the official vuze wiki, but doesn't exits on the new BiglyBT wiki.

What is `Constants.AZUREUS_WIKI` ? The current implementation references to https://wiki.biglybt.com/w/NAT_problem
which doesn't exists. Is that constant deprecated or is this instance supposed to link to https://wiki.vuze.com/w/NAT_problem ?

TuxPaper:
After talking about it with the team, I'm afraid we won't be able to accept this pull unless the wiki urls are ``public final static String``.  For example:
```
public final static String URL_WIKI_FOOBAR = Constants.URL_WIKI + "foobar";
```

The main reason for this is that Java will inline compile-time constants.  The enum approach builds a static list of objects, and requires string appending to get the full URL string. While that's not a huge footprint issue, the advantage of inlining is a bigger positive.

uis/src/com/biglybt/ui/swt/mainwindow/MainStatusBar.java
TuxPaper: Due to legal risks that we aren't willing to take, we try not to link to any vuze.com URLs.  The fact that the Google results brings up vuze.com URLs is completely out of our control :)

We hope that over time, people would write new wiki pages specifically for BiglyBT.

uis/src/com/biglybt/ui/swt/devices/DevicesFTUX.java
TuxPaper: Probably :)  I'm sure there's also other places that aren't using LinkLabel.
I have some pending changes to the ConfigSection classes that fix a few other non-LinkLabel usage

uis/src/com/biglybt/plugin/startstoprules/defaultplugin/ui/swt/ConfigSectionDownloading.java
TuxPaper: This class will be gone/completely restructured soon.  Probably in a few days after we are sure the release went ok.

TuxPaper:
One other note about the URL class.. it does DNS lookups and other crap when comparing them to other URLs.  For this reason, we end up just avoiding the URL class until it's actually needed (if needed at all).

uis/src/com/biglybt/plugin/startstoprules/defaultplugin/ui/swt/ConfigSectionDownloading.java
lislei: Sweet - I'll get back to it when the changes are ready 👍

uis/src/com/biglybt/ui/swt/mainwindow/MainStatusBar.java
lislei: This field should be `@Depricated` then and possibly removed. 
I'm not going to suggest removing it since it's part of the public api and i have no idea about it's usage. 
- Is there any repository or project that has all the different plugins linked in as source code?

lislei:
Are you refering to Utils.launch(URL)? 
I assumed it took advantage of the url object, but you're right - it's converts it back to a String representation... Sure, I'll make some optimalizations to it.

lislei:
I made some improvements so everything is computed at compile time. 
I'll argue the number of enums are limited and thereby has a neglectible overhead. The enum pattern has some flexible traits, being easily extendable when needed is one of them. 

Anyhow you slice it, not looking up the url value from the resource bundle is an major improvement! This was done in quite a few places.

Please let me know what you think.  :)

uis/src/com/biglybt/ui/swt/mainwindow/MainStatusBar.java
TuxPaper: Deprecated definitely makes sense.  I'm just guessing, but the reason it's probably still around is it was just easier to change the constant value than to replace the constant name.  That, and I think the "w/" is what makes our Google Search redirect work.

TuxPaper:
Not quite. What I was saying is remove the enum and replace it with several `public final static String` lines.  This will also remove the need to call "toString()" everywhere it's used, making the code look cleaner.  So, the first two would be

```
public static final String ADVANCED_NETWORK_SETTINGS = URL_WIKI + "w/UG_Options#Advanced_Network_Settings";
public static final String ALL_TRACKERS_VIEW = URL_WIKI + "w/AllTrackersView";
```
(variable names are up to you, but for constants, they are usually all caps)

I agree, at the end of the day, just having the URLs out of the messagebundle is a very good thing.

lislei:
Excellent, I see we're heading for down to bare bone performance. Haven't practised that in a while. Will be fun! :)

TuxPaper:
Yes :) Keep It Simple :)  As you probably realized, our code base is large, interconnected, and confusing, so the more we can do to simplify it, the better.

core/src/com/biglybt/core/internat/MessageText.java
```diff
@@ -110,6 +110,30 @@
...
+	  CONSTANTS.put("Alert.failed.update.url", Wiki.FAILED_UPDATE);
+	  CONSTANTS.put("alltrackers.link.url", Wiki.ALL_TRACKERS_VIEW);
+	  CONSTANTS.put("azbuddy.classic.link.url", Wiki.FRIENDS);
+	  CONSTANTS.put("azbuddy.dchat.link.url", Wiki.DECENTRALIZED_CHAT);
+	  CONSTANTS.put("restart.error.url", Wiki.RESTARTING_ISSUES);
+	  CONSTANTS.put("unix.script.new.manual.url", Wiki.UNIX_STARTUP_SCRIPT);
+	  CONSTANTS.put("wiki.fat32", Wiki.FAT32_FILE_SIZE_LIMIT);
+	  CONSTANTS.put("url.wiki.app.disappears", Wiki.APPLICATION_DISSAPEARS);
+	  CONSTANTS.put("url.wiki.failed.update", Wiki.FAILED_UPDATE);
+	  CONSTANTS.put("url.wiki.swt.cant.autoupdate", Wiki.SWT_CANT_AUTO_UPDATE);
```
lislei: I'm starting to believe that expansion keys for wiki urls could be eliminated as well. 
This would require some rewrite of the text. 
- It could seem doable with that POC for updating crowdin translations. 

For this particular key there it is already parameterized. One could simply add `%4` together with the `Wiki.`constant

core/src/com/biglybt/plugin/net/buddy/BuddyPlugin.java
```diff
@@ -288,11 +289,10 @@
 					});
 	
 			os_param.setVisible( SUPPORT_ONLINE_STATUS  ); // If we add this then use proper message texts in the STATUS_STRINGS
-	
-			
-			HyperlinkParameter profile_link = 
+
+			HyperlinkParameter profile_link =
 					config.addHyperlinkParameter2( 
-						"azbuddy.profile.info", MessageText.getString( "azbuddy.profile.info.url" ));
```
lislei: @parg @TuxPaper 
Is this a safe change? 
I can't determine if the different plugins are supposed to use a spesific API or not. 
Is there a difference between plugins shipped in "core" module and the external ones?

lislei: If there is a strict API then a forward compatible change would probably be
`Constants.URL_WIKI + "w/Friends#public-profile"`

parg: Not sure if your changes handle this - these two keys

azbuddy.send.friend.key.msg
azbuddy.add.friend.key.msg

are sent via chat unexpanded to explain what's going on to the recipient. On the receiver's side they are expanded and shown to the user (in their local language of course). Both of these message texts contain a wiki url expansion:  {azbuddy.classic.link.url}

As for plugin stuff, they are supposed to use the plugin interface's LocaleUtilities interface and not MessageText. However, for built-in ones the need to stick to plugin-interface stuff is widely flouted. Indeed, for a lot of non-built-in ones (particularly ones written by ourselves), non-PI stuff is often used. Given that we hardly have a vibrant community of plugin developers I really wouldn't consider formalizing plugin access to things, just try and ensure that things don't get broken :)

lislei: Thank you for elaborating. 
I believe the keys you mention won't pose any issue as it's up to the receiver to translate them. 

I'm moving the expansion keys needed for this plugin to the initialization block - as they are spesific to this plugin only.
Like you're saying, as long as it's a built in plugin then there's no need to be too worried about strictly conforming to the plugin-api.

uis/src/com/biglybt/ui/swt/mainwindow/MainStatusBar.java
lislei: If someone could guarantee it's safe to delete it I could. That would be cleaner....

parg: It should definitely not be deleted - when a page doesn't exist (yet) on the BiglyBT wiki we show a general google search result for the terms involved. This quite often leads to the top hit being quite relevant - open https://wiki.biglybt.com/w/NAT_problem in google for example.

lislei: Sorry, I realize I wasn't spesific enough. My bad. 

What I believe is the `Constants.AZUREUS_WIKI` to be a duplicate of `Constants.URL_WIKI`. The only difference between those two is the first has the "w/" path built in. @parg , is there any reason to hold on to the `AZUREUS_WIKI` constant for plugin compatibility?

parg: Ah, my misunderstanding! 

I have a large number of the 'real' plugins loaded in Eclipse and none of them refer to AZUREUS_WIKI - so go for it and remove it, I'll take the flak if it breaks anything!

TuxPaper:
Ok, my ui parameter stuff is now committed, and I totally broke your pull :/ Whee!

lislei:
I like the clean look of it. I even noticed a decrease in overall file size on that particular beta release. *thumbs up*

uis/src/com/biglybt/ui/swt/mainwindow/MainStatusBar.java
lislei: I added a commit removing the AZUREUS_WIKI constant.

core/src/com/biglybt/core/networkmanager/impl/tcp/SelectorGuard.java
```diff
@@ -117,8 +117,9 @@ public void verifySelectorIntegrity( int num_keys_ready, long time_threshold ) {
         //Under windows, it seems that selector spin can sometimes appear when >63 socket channels are registered with a selector
```
lislei: This relic if-block is always skipped and should be removed.

The implications of removing it is removing the only usage of 
`listener.safeModeSelectEnabled()` and `listener.spinDetected()`.
@parg, this is your domain. Please advice.

core/src/com/biglybt/core/networkmanager/impl/tcp/SelectorGuard.java
```diff
@@ -137,8 +138,9 @@ public void verifySelectorIntegrity( int num_keys_ready, long time_threshold ) {
 
 
     if( consecutiveZeroSelects > SELECTOR_FAILURE_THRESHOLD ) {  //should only happen under Windows + JRE 1.4
```
lislei: Is ther a flaw in the semantics here preventing this if-block ever to be reached?
The previous if-block takes precedence. Would make sense if they switched.

core/src/com/biglybt/core/networkmanager/impl/tcp/SelectorGuard.java
parg: Yup, looks that way

core/src/com/biglybt/core/networkmanager/impl/tcp/SelectorGuard.java
parg: If the block is never reached given current code paths then go ahead and remove it (and check implications of failureDetected() now never being called if you want)

core/src/com/biglybt/core/networkmanager/impl/tcp/SelectorGuard.java
lislei: I opened #1005 for this

core/src/com/biglybt/core/networkmanager/impl/tcp/SelectorGuard.java
lislei: See #1005

TuxPaper:
Still a WIP?

Looks ready to merge to me :)

lislei:
It's time to merge it. 
This PR removes all the single substitution keys for wiki urls from the message bundles. 
This removes superflous keys from Crowdin and urls being accidentaly translated.

There are still some urls left in the message bundles that are hardcoded in there together in the translations. Id like to have a second PR dealing with those.

TuxPaper:
Thanks for the iterations and cleanup effort.  Much appreciated.

PR #964, opened 2019-03-25
Codebase encoding

lislei:
I'm creating this PR as a test and a place for discussion.

I ran a search for CRLF line encodings and it trew me an error searching trough the files. 
The tool used pointed out that [this file](https://raw.githubusercontent.com/BiglySoftware/BiglyBT/master/core/src/com/biglybt/core/metasearch/impl/DateParserRegex.java) had some illegal characters for the windows codepage. 
If so then that's true for the ISO-8859-1 as they are fairly similar..

There are several interesting aspects to this. One is that github works great with UTF8... :)

### DIFF
--- a/core/src/com/biglybt/core/metasearch/impl/DateParserRegex.java
+++ b/core/src/com/biglybt/core/metasearch/impl/DateParserRegex.java
@@ -503,8 +503,8 @@ public class DateParserRegex extends DateParser {
 		dateParser.parseDate("20-Dec-07");	//
 		dateParser.parseDate("2009-01-12 at 03:36:38" );
 		dateParser.parseDate("2013-08-11T18:30:00.000Z" );
-		dateParser.parseDate("12小时前");
-		dateParser.parseDate("12 小时前");
+		dateParser.parseDate("12\u5C0F\u65F6\u524D");
+		dateParser.parseDate("12 \u5C0F\u65F6\u524D");
 	}
 
 


lislei:
What is the right encoding for source files? 
My IntelliJ integration tells me 8859_1 encoding for all java files except for this one. 
I always assumed that to be right but now im not sure anymore.
 
The `maven-compiler-plugin` lacks `<encoding>` tag for defining the encoding for '.java' files. It would be an improvement to specify it!

The `<project.build.sourceEncoding>8859_1</project.build.sourceEncoding>` is for copying resource files  - especially the l10n.properties files. That shows up right in IntelliJ at my end.

BiglyBT:
I honestly don't know what the proper encoding for java files are!  The "\uXXXX" approach at least makes it so we have no special characters, and thus utf-8 and 8859_1 compatible

TuxPaper:
I honestly don't know what the proper encoding for java files are! The "\uXXXX" approach at least makes it so we have no special characters, and thus utf-8 and 8859_1 compatible

lislei:
If utf8 then these special characters found are legal and not a problem after all..

I am thinking if the build and ship process is on Linux/Mac and system property `file.encoding` isn't set then utf8 is what it should be. If not then the first 256 characers are the same in both utf8 and iso-8859-1.

My advice is to define the source encoding - that would undoubtedly be an improvement removing any doubt :)

PR #965, opened 2019-03-25
Normalized line endings

lislei:
Defined LF line endings in .gitattributes

Normalized line endings by following [this procedure](https://help.github.com/en/articles/dealing-with-line-endings).
This converted all CR+LF line endings to LF.

Added `.gitattributes`. Please add any missing file extensions and mark them either as binary or text.


TuxPaper:
The CRLF annoy me too, and I really wish I would have converted them when we branched Vuze.

But now, I don't want to overwrite the blame history.  Although I'm willing to bet 95% of the files have no BiglyBT blame history other than "Initial Commit".

What I may do at some point (motivation is low at the moment) is determine which files have no blame history, and change them to LF using the BiglyBT user.  We try to use the BiglyBT user as the "don't blame anyone!" user :P

lislei:
Turns out that all the initial files from Vuze have unix style LF line feeds. 
All the files in this commit are new files produced after the initial commit.

I suggest that as a minimum the .gitattributes file should be merged into master to limit any new instances.

ferdnyc:
@lislei 
> I suggest that as a minimum the .gitattributes file should be merged into master to limit any new instances.

Is that safe to use around existing non-conforming files? Or will editing any of the files currently in  CRLF format cause the entire file to be converted in the commit for that edit? (Which would have the same side-effect @TuxPaper was concerned about, where that one commit steamrolls over the entire blame history.) ISTR hitting problems with conversion side-effects in a different project, after a `.gitattributes` with line-ending configs was committed.

lislei:
Excellent question @ferdnyc.
Your statement is correct. Existing files would be be converted when commiting changes to any CRLF file.
This configuration alone might be inadequate.

Options that I am aware of are:
* Identify affected files and check them in with LF.
** Could be original author, or a system user (in this case BiglyBT). 
* Rewrite commits with CRLF line endings and do a forced push on your master branch.
** See filter-branch https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History
* Don't declare line endings in `.gitattribute`
** Developers with `core.autocrlf` attributes set wil get aquintet with "the steamroller" as mentioned abowe. This is by the way the reccomanded github setting: https://help.github.com/en/articles/dealing-with-line-endings

lislei:
On another note - I've recently discovered that `git blame` command has some switches that ignores whitespace and eaven if files are moved around. 
IntelliJ has adopted those settings as default. When i "annotate" any file it ignores whitespace changes.
I'm predicting it's only a matter of time before github adopts this feature making the blame history cleaner.

TuxPaper:
Github has a ?w=1 url parameter that works on some of their pages, but not all.  Ignoring whitespace is definitely one of my favorite features.  I hope Github makes it default someday.

ferdnyc:
I also discovered this in my travels: [git-hyper-blame](https://commondatastorage.googleapis.com/chrome-infra-docs/flat/depot_tools/docs/html/git-hyper-blame.html).
> <h4>Description</h4>
>
> `git hyper-blame` is like `git blame` but it can ignore or "look through" a given set of commits, to find the real culprit.
> 
> This is useful if you have a commit that makes sweeping changes that are unlikely to be what you are looking for in a blame, such as mass reformatting or renaming. By adding these commits to the hyper-blame ignore list, `git hyper-blame` will look past these commits to find the previous commit that touched a given line.
> 
> Follows the normal `blame` syntax: annotates `<file>` with the revision that last modified each line. Optional `<rev>` specifies the revision of `<file>` to start from.
> 
> Automatically looks for a file called `.git-blame-ignore-revs` in the repository root directory. This file has the same syntax as the `--ignore-file` argument, and any commits mentioned in this file are added to the ignore list.

But, of course, since it's an add-on command it requires that everyone using the repo have it installed, and it's no use with GitHub or other service providers' offerings.

ferdnyc:
What I _have_ done that worked pretty well to limit **new** instances of CRLF files being created was, rather than using `.gitattributes` to enforce line endings, I installed an [`.editorconfig`](https://editorconfig.org) that configured default parameters like:
```ini
[*]
end_of_line = lf
insert_final_newline = true # (HUUUGE pet peeve for me!)
```
And then just made explicit `end_of_line = crlf` exceptions for all of the _existing_ CRLF files.

Looking through my development directory right now, it seems to have become surprisingly common for open-source code repositories to distribute their own top-level `.editorconfig` files. I never even realized most of these were there!
```console
$ locate .editorconfig|wc -l                                    
124
```

TuxPaper:
editorconfig has been on my list of "look into" for a long time now.  It would be awesome someday to have an editorconfig that links to either a pargStyle or a tuxStyle, where appropriate.  Along with "reformat before commit", and we could all code in the style we want and commit in the proper format.  But editorconfig seems very basic in that regards.  Still, tab info would go a long way.

Related, I wouldn't doubt it might be possible to format the java code with Intellij's formatter now, instead of using the Eclipse plugin.  When I first started using Intellij, there were 3 or 4 things that IntelliJ couldn't duplicate from Eclipse's formatter.  Surely it can now.

PR #978, opened 2019-04-02
code cleanup: Removed "single-buffer mode" fallback mode 

lislei:
This was a fallback/compatibility mode for Java versions prior to Java 5

### DIFF
--- a/core/src/com/biglybt/core/networkmanager/impl/tcp/TCPTransportHelper.java
+++ b/core/src/com/biglybt/core/networkmanager/impl/tcp/TCPTransportHelper.java
@@ -28,14 +28,11 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Random;
 
-import com.biglybt.core.logging.LogAlert;
-import com.biglybt.core.logging.Logger;
 import com.biglybt.core.networkmanager.VirtualChannelSelector;
 import com.biglybt.core.networkmanager.VirtualChannelSelector.VirtualSelectorListener;
 import com.biglybt.core.networkmanager.impl.TransportHelper;
 import com.biglybt.core.proxy.AEProxyAddressMapper;
 import com.biglybt.core.proxy.AEProxyFactory;
-import com.biglybt.core.util.Constants;
 import com.biglybt.core.util.Debug;
 import com.biglybt.core.util.TimeFormatter;
 
@@ -57,8 +54,6 @@ TCPTransportHelper
 
 	private long remainingBytesToScatter = 0;
 
-	private static boolean enable_efficient_io = !Constants.JAVA_VERSION.startsWith("1.4");
-
 	private	final SocketChannel	channel;
 
 	private ByteBuffer	delayed_write;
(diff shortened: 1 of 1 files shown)


lislei:
The property `network.tcp.enable_safe_selector_mode` 
was always removed or disabled when running java version > 5

parg:
You removed too much code

if ( enable_efficient_io && remainingBytesToScatter < 1 ){
...

There's an '&&' in there...

Also there's a bug in the tcp_address optimization code there, I would think that the correct fix is to remove the InetSocketAddress in

	    InetSocketAddress	tcp_address = applied_mapping.getRemoteAddress();

core/src/com/biglybt/core/networkmanager/impl/tcp/TCPTransportHelper.java
```diff
@@ -408,40 +361,7 @@ private int channelWrite(ByteBuffer buf) throws IOException
...
-
-				bytes_read += read;
-
-				if( read < data_length ) {
-
-					break;
-				}
-			}
-		}
+		bytes_read = channel.read(buffers, array_offset, length);
```
lislei: Are the read and write strategies symmetrical? Here they didn't end up using `single-buffer mode`.

core/src/com/biglybt/core/networkmanager/impl/tcp/TCPTransportHelper.java
```diff
@@ -285,43 +274,7 @@ public TCPTransportHelper( SocketChannel _channel ) {
...
-				}
-			}else{
-
-					//single-buffer mode
-
-				for( int i=array_offset; i < (array_offset + length); i++ ) {
-
-					int data_length = buffers[ i ].remaining();
-
-					int written = channelWrite(buffers[i]);
```
lislei: `channelWrite()` is the only place that modifies `remainingBytesToScatter`. When using efficient_io mode then this was left unaltered by this class. 
Only other entry point is thru `public void setScatteringMode(long forBytes)`.

My first imression is that if you're in `enable_efficiant_io` mode then you're bound to it.

core/src/com/biglybt/core/networkmanager/impl/tcp/TCPTransportHelper.java
parg: While you perhaps see this as an interesting voyage of archaeological discovery, I see this code as a heavily overgrown battlefield where I lost many comrades. 

There is a massive amount of crap buried which may or may not now be ready for disposal, the danger lies in the costs of removing something that appears to be 'no longer required' which, when long forgotten and released in 3 months time, causes random application hangs resulting in me (not you) having to invest life-span reducing effort into diagnosing. 

That selector-spin issue took literally weeks, maybe even months, to get into a state the behaved reasonably well on all platforms (follow the tendrils of the spin detection through the codebase if you have a spare moment or two). Of course we were using NIO at the start of its life, when the JVM was still full of bugs. Perhaps they are all fixed now, perhaps not... The safe-selector-mode is currently designed to auto-disable itself every startup in the hope that one day it will not be required, however we're not quite there yet!

Don't get me wrong, I love that you're spending time looking into removing redundant code, just please be sure that the code really is 100% dead. 

Regarding single-buffer-mode, I don't know the code but I do know that whatever it is it currently works, and that's good enough for me.

PR #979, opened 2019-04-02
code cleanup: removed "safe_selector_mode" compatibility mode 

lislei:
See javadoc on "SelectorGuard". This issue was fixed in Java 5.


parg:
What makes you think that safe-selector-mode is no longer required? Regardless of whatever javadoc you might have read I've seen it being enabled when spinning is detected (on OSX) via enableSafeSelectionMode()

Please don't get carried away removing stuff because you believe it is no longer required, I don't have time to go checking everything so unless it is a very obvious consequence of something, leave it alone to fester.

core/src/com/biglybt/core/config/impl/ConfigurationChecker.java
```diff
@@ -384,17 +384,6 @@
...
 
-	    	if // disable safe selector mode enabled at some point in the past if we're on java 6 or higher and/or not on windows
-	    	(	COConfigurationManager.getBooleanParameter("network.tcp.enable_safe_selector_mode")
-	    		&& !(Constants.isWindows &&
-	    			(Constants.JAVA_VERSION.startsWith("1.4") ||
-	    			Constants.JAVA_VERSION.startsWith("1.5"))
-	    		)
-	    	)
-	    	{
-	    		COConfigurationManager.removeParameter("network.tcp.enable_safe_selector_mode");
```
lislei: I'm assuming an entry point from static initilization to this block of code. 

If I read this right it removes any configured value for `network.tcp.enable_safe_selector_mode=true`
when not running windows under java 1.4 and not running windows under java 1.5.

Is there anything that might hook this back in?

My point is to eliminate java version checks for java versions not longer supported.

lislei: I believe I found a entry point that you mentioned that enables all this after all. Thanks for pointing that out. 

I guess this code block effectively wipes the persistent value `network.tcp.enable_safe_selector_mode` for all run configurations. The if-block could be simplified to one line of code.

PR #995, opened 2019-04-12
code cleanup: removed duplicate if-conditions

lislei:
Running an inspection found two cases of duplicate if conditions.


TuxPaper:
heh, for a moment I was worried this pull was going to be a billion lines of code :)

(For the record, unless there's a major bug, changing a billion lines of code would be frowned upon :P )

PR #996, opened 2019-04-12
Some minor optimalizations

lislei:
Collections.singletonList doesn't have the array creation and performs better.

### DIFF
--- a/core/src/com/biglybt/core/devices/impl/DeviceImpl.java
+++ b/core/src/com/biglybt/core/devices/impl/DeviceImpl.java
@@ -547,7 +547,7 @@ DeviceImpl
 			return( Collections.EMPTY_LIST );
 		}
 		
-		return( Arrays.asList( image_id ));
+		return Collections.singletonList(image_id);
 	}
 	
 	@Override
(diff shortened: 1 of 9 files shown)


TuxPaper:
I'm surprised Android has Collections.singletonList! (It does since forever)

Did you check every place the resulting List is used to see if it assumes the list is an ArrayList or tries to add or remove from the list?  I get this dirty feeling that we assume all over the place that Lists are modifiable.

lislei:
Cases improved are all replacing [Arrays.asList](https://developer.android.com/reference/java/util/Arrays.html#asList(T...)).

This gives you a fixed list implementation that doesn't allow add or remove operations. Note that this implementation is an inner class if `Arrays`.

I checked everywhere for potential `set(idx, obj)` and `sort(Comparator)` bur found only read access.

TuxPaper:
> This gives you a fixed list implementation that doesn't allow add or remove operations

Yes, that's exactly why I asked if you checked the calling code to see if it tried to add :)  Arrays.asList() returned an ArrayList which calling code could have modified.

Thanks for checking the callers.  It's literally the longest part of the process, because sometimes variables get passed all over the place and tracing their call stack can often be very time consuming.

One future request regarding imports.  If you can set your `File > Settings > Project Settings > Code Style > Java > Imports > General > Class count to use import with '*'` to a much lower number, future diffs will be smaller. 4 seems to be common in our code, but I've seen 5 as well.

PR #998, opened 2019-04-13
performance: Bulk loading of collections

lislei:
Optimized adding to a collections with bulk loading methods.
This shortens and cleans up the code.

[Collections.addAll(Collection, vararray)](https://developer.android.com/reference/java/util/Collections.html) has been around for a while and is andriod safe.


parg:
While these changes all appear to be correct, please ensure that if you create pull requests for optimizations then you really are making a quantifiable improvement as checking things takes time. 

For example, if you find a code path that is executed multiple times a second then it is worth digging into. If once a minute then unless there is a massive performance gain it isn't worth it. 

Possibly you're already doing this - if so then please document the request with justification. Cheers!

lislei:
Yup, you're right. The naming is a bit off. 
I constantly find myself improving JIRA titles at work...

Is it too late to tweak the title of this PR?

PR #999, opened 2019-04-15
Relabel "Set Preferred Width" for table columns to "Size Column to Fit"

ferdnyc:
Selecting "Set Preferred Width" for a column doesn't STORE the column width at all —  as it's implemented, it sets the column width _to_ its preferred width:

https://github.com/BiglySoftware/BiglyBT/blob/0665dbab0bc0a2a0bdb37f1f180ab9caf19c73bc/uis/src/com/biglybt/ui/swt/views/table/impl/TableViewSWT_Common.java#L1212-L1226

From the user's perspective, "Reset Width" seems a better description of the action.


TuxPaper:
You've found a can of bugs! haha.

Story time!
So, technically the preferred width feature is supposed to calculate the width the column wants to be to display all the text.  It was designed so we could double click on the table headers in between the columns, and have the column expand to show all the text.  This feature is in Windows file Explorer, and is labeled "Size Column to Fit"

"Size Column to Fit" seems like the most appropriate name change, as it would be familiar to windows users.  "Reset Width" option would be more of a "set back to original width specified by plugin" which may or may not cut off the cell's text.  Right now the only way to reset width is to "Reset Columns" which does way more than just reset the width (it resets which columns are displayed)

So, as for the can of bugs.. we never implemented (or it got removed!?) the double click between column headers! :(  I'm sad.
Also, the "Set Preffered Width" doesn't play very nice with some columns, like the Rating column, which shrinks to almost 0 px! HA

ferdnyc:
@TuxPaper I'll buy that. Really my main concern is to get the word "set" out of there, since like I said... 

"Resize to contents", "Fit to contents" (and more variations) is another form I've seen. The GIMP calls that menu selection for its window sizing "Shrink Wrap", because... well... it's the GIMP. (More "cautionary tale" than "example to emulate", let's just say.)

I pushed in another commit to change it to "Size Column to Fit".

> You've found a can of bugs! haha.

(I thought that can sounded funny when I shook it, but I still had to look...)

You don't know the half of it, TBH. While I had the code loaded in, I made the mistake of letting IntellijIDEA run its code-cleanup scan, and then started perusing the deprecation notices... long story short, I finally reached the sensible conclusion that down that road lie madness, but that's 45 minutes of my life I'll never get back. :grin: 

But it was mostly just procrastination on all the _other_ unpulled threads in my day, since I'm in the throes of one of those things where I've got like 60-70 tabs open across four different browser windows, some of them are **_hours_** old, and so far I think I've _started_ like 17 different things, in total, and completed exactly none of them.

So, hey, don't feel so bad about the column-divider double clicking stuff.

(P.S> Did you know that `getOffsetAtLocation(Point)` is [deprecated in SWT](https://help.eclipse.org/2019-03/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/swt/custom/StyledText.html#getOffsetAtLocation(org.eclipse.swt.graphics.Point)) in favor of the functionally-equivalent `getOffsetAtPoint(Point)`, because "it does not throw an exception when no character is found and thus performs faster." **#TheMoreYouKnow**)

TuxPaper:
This will be an interesting test of CrowdIn as well.  I saw it recently take one of my English edits and leave the other languages intact.  Previously, it would wipe out they translation for that key for all the other languages.

TuxPaper:
Well, that's interesting, Crowdin kept the other translations.

Pro: At least there's a translation
Con: Translation doesn't reflect new wording

Next step for Crowdin would be to mark that key as "unapproved" so it would pop up again for proofreaders

ferdnyc:
> Well, that's interesting, Crowdin kept the other translations.

Yet was still nice enough to supply 56 new commits, to insert the blank line at the bottom of every translated MessageBundle that resulted because I had IDEA repair the missing newline at the end of the English master.

"Helping!™" 🙄

PR #1005, opened 2019-04-15
Code cleanup: removed relic SelectorGuard handling

lislei:
The "safe selector mode" was a workaround for Java 1.4/1.5 under Windows.

When this particular workaround detected a spin, then and only then the property `"network.tcp.enable_safe_selector_mode"` would be set to `true`.

The above property was reset to `false` after a start/restart when the combo (operating system + java version) was different then for this workaround.

I'm leaving this as a work in progres (wip) for discussion.

### DIFF
--- a/core/src/com/biglybt/core/networkmanager/VirtualChannelSelector.java
+++ b/core/src/com/biglybt/core/networkmanager/VirtualChannelSelector.java
@@ -23,44 +23,22 @@ import java.nio.channels.SelectionKey;
 import java.nio.channels.ServerSocketChannel;
 import java.nio.channels.SocketChannel;
 import java.nio.channels.spi.AbstractSelectableChannel;
-import java.util.*;
 
-import com.biglybt.core.config.COConfigurationManager;
-import com.biglybt.core.logging.LogEvent;
-import com.biglybt.core.logging.LogIDs;
-import com.biglybt.core.logging.Logger;
 import com.biglybt.core.networkmanager.impl.tcp.VirtualChannelSelectorImpl;
-import com.biglybt.core.util.AEMonitor;
-import com.biglybt.core.util.Debug;
 
 
 public class VirtualChannelSelector {
-  private static final LogIDs LOGID = LogIDs.NWMAN;
   public static final int OP_ACCEPT  = SelectionKey.OP_ACCEPT;
   public static final int OP_CONNECT  = SelectionKey.OP_CONNECT;
   public static final int OP_READ   = SelectionKey.OP_READ;
   public static final int OP_WRITE  = SelectionKey.OP_WRITE;
 
-  private boolean SAFE_SELECTOR_MODE_ENABLED = TEST_SAFE_MODE;
-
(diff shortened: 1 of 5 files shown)


core/src/com/biglybt/core/networkmanager/VirtualChannelSelector.java
```diff
@@ -41,7 +41,7 @@
   public static final int OP_READ   = SelectionKey.OP_READ;
   public static final int OP_WRITE  = SelectionKey.OP_WRITE;
 
-  private boolean SAFE_SELECTOR_MODE_ENABLED = TEST_SAFE_MODE || COConfigurationManager.getBooleanParameter( "network.tcp.enable_safe_selector_mode" );
+  private boolean SAFE_SELECTOR_MODE_ENABLED = TEST_SAFE_MODE;
```
lislei: There isn't any way to enable "safe selector mode" anymore. 
I don't believe there's any valid reason to keep the option to test it.. is there?

core/src/com/biglybt/core/networkmanager/VirtualChannelSelector.java
```diff
@@ -422,14 +422,6 @@ public void destroy()
 
   public boolean isSafeSelectionModeEnabled() {  return SAFE_SELECTOR_MODE_ENABLED;  }
 
-  public void enableSafeSelectionMode() {
-    if( !SAFE_SELECTOR_MODE_ENABLED ) {
-      SAFE_SELECTOR_MODE_ENABLED = true;
-      COConfigurationManager.setParameter( "network.tcp.enable_safe_selector_mode", true );
-      initSafeMode();
```
lislei: initSafeMode() is only called from the constructor:
```
    if( SAFE_SELECTOR_MODE_ENABLED ) {
      initSafeMode();
    }
```

There is also some if switches in the implementing methods like:
```
  public boolean isPaused( AbstractSelectableChannel channel ) {
	    if( SAFE_SELECTOR_MODE_ENABLED ) {
              ...
	    }
	    else {
	      return( selector_impl.isPaused( channel ));
	    }
  }
```

@parg: What do you suggest to be done with all this?

lislei:
As for removing spinDetected() and failureDetected() makes the whole interface GuardListener obsolete.

core/src/com/biglybt/core/networkmanager/VirtualChannelSelector.java
parg: If the code has become unreachable purely from following through things that are logical tautologies then go ahead and remove them. Just don't remove stuff based on vague comments or things that appear to be true but can't be proved as such :)

lislei: Well served. Haven't heard that word in quite a while :)

lislei:
I removed the loose ends from "safe selector mode" and ended up removing another 250+ lines of code. 
There must have been some serious battles fought with connections and cpu hogging. 
If you ever need a topic for your new book, then you might find inspiration from these two commmits.

PR #1011, opened 2019-04-19
code cleanup: Bulk loading of collections

lislei:
I like the expression "less is more" for source code. 
This combined with a more optimized usage of the Java API will improve the overall maintainability.

This one slipped away from one of the last PRs :)


parg:
for (int j=0;j<peers.length;j++){

					result.add( peers[i] );
				}

hmmmmmmm

parg:
You want 'less is more' - I deleted all the cache discovery code :)

PR #1026, opened 2019-04-23
Code cleanup: mitigated UnsupportedEncodingException for String creations

lislei:
Eliminated try-catch constructs for new String(byte[], (String)charsetname) creations.

I've only changed the simple cases. Other cases of try-catch(Throwable/IOException) with multiple sources of exceptions are left untouched.

### DIFF
--- a/core/src/com/biglybt/core/history/impl/DownloadHistoryManagerImpl.java
+++ b/core/src/com/biglybt/core/history/impl/DownloadHistoryManagerImpl.java
@@ -770,29 +770,24 @@ DownloadHistoryManagerImpl
 
 		// System.out.println( "loading " + file );
 
-		try{
-			if ( FileUtil.resilientConfigFileExists( file )){
+		if (FileUtil.resilientConfigFileExists(file)) {
+			Map map = FileUtil.readResilientConfigFile(file);
 
-				Map map = FileUtil.readResilientConfigFile( file );
+			List<Map<String, Object>> list = (List<Map<String, Object>>) map.get("records");
+			if (list != null) {
 
-				List<Map<String,Object>>	list = (List<Map<String,Object>>)map.get( "records" );
+				for (Map<String, Object> m : list) {
 
-				for ( Map<String,Object> m: list ){
+					try {
+						DownloadHistoryImpl record = new DownloadHistoryImpl(result, m);
+						result.put(record.getUID(), record);
 
-					try{
-						DownloadHistoryImpl record = new DownloadHistoryImpl( result, m );
+					} catch (Throwable e) {
 
-						result.put( record.getUID(), record );
-
-					}catch( Throwable e ){
-
-						Debug.out( e );
+						Debug.out(e);
 					}
 				}
 			}
-		}catch( Throwable e ){
-
-			Debug.out( e );
 		}
 
 		return( result );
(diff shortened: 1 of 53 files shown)


core/src/com/biglybt/core/history/impl/DownloadHistoryManagerImpl.java
```diff
@@ -770,29 +769,17 @@ public void perform(TimerEvent event) {
 
 		// System.out.println( "loading " + file );
 
-		try{
-			if ( FileUtil.resilientConfigFileExists( file )){
-
-				Map map = FileUtil.readResilientConfigFile( file );
```
lislei: I believe this is the instance you are referring to. 
These two FileUtil methods seems to have a failsafe implementation. The javadoc states so as well:
```
	/**
	 *
 	 * @return Map read from config file, or empty HashMap if error
	 */
```

core/src/com/biglybt/core/history/impl/DownloadHistoryManagerImpl.java
```diff
@@ -917,34 +897,22 @@ public void perform(TimerEvent event) {
...
-				complete_time 	= (Long)map.get( "c" );
-				remove_time 	= (Long)map.get( "r" );
+			Long l_size = (Long) map.get("z");
 
-			}catch( IOException e ){
+			size = l_size == null ? 0 : l_size;
 
-				throw( e );
-
-			}catch( Throwable e ){
```
lislei: Thanks, I see what you mean now.. 
I'll re-examine the other changes!

core/src/com/biglybt/core/util/UrlUtils.java
```diff
@@ -1254,36 +1251,31 @@ public static String unescapeXML(String s) {
 
 		Map	result = new HashMap();
 
-		try{
```
lislei: Variable `headers_to_use` is always none null, so no exceptions from this block.

lislei:
Sorry for not spotting just that. I've fixed up the changes to make it behave just like before. That's the intent anyways.

lislei:
Umm.. I've encoutered this construct alot:
```
byte[]	bytes = (byte[])obj;

try{
    return( new String( bytes, "UTF-8" ));
}catch( Throwable e ){
    return( new String( bytes ));
}
```
Researching and testing the String constructor I've found that it either throws NullpointerException or UnsupportedEncodingException with JDK8. This behaviour might have changed over the last versions of java. 
The JDK5, JDK8 and Andriod documentation informs that;

> The behavior of this constructor when the given bytes are not valid in the given charset is unspecified. The CharsetDecoder class should be used when more control over the decoding process is required.

core/src/com/biglybt/core/history/impl/DownloadHistoryManagerImpl.java
```diff
@@ -770,29 +770,24 @@ public void perform(TimerEvent event) {
...
-						DownloadHistoryImpl record = new DownloadHistoryImpl( result, m );
+				for (Map<String, Object> m : list) {
 
-						result.put( record.getUID(), record );
+					try {
+						DownloadHistoryImpl record = new DownloadHistoryImpl(result, m);
+						result.put(record.getUID(), record);
 
-					}catch( Throwable e ){
+					} catch (Throwable e) {
```
lislei: Since the FileUtil methods are fail safe then this try-catch block should now be sufficiant enough...

PR #1028, opened 2019-04-24
SWT: Only use setBackgroundMode(SWT.INHERIT_FORCE) on Win

### LINKED ISSUE #1002: Labels of some buttons are unreadable on linux
OS and version linux, Arch Linux
For *nix: Desktop Environment KDE Plasma 5.15.4
Java 10.0.2 (64 bit)
Oracle Corporation
/usr/lib/jvm/java-10-openjdk

SWT v4922, gtk/3.24.8, zoom=100, dpi=96
Linux v5.0.7-arch1-1-ARCH, amd64 (64 bit)
B1.9.0.0/4 az3

Some buttons are unreadable on linux.
See the screenshots below.

biglybt5

"All" button below the search field is unreadable.
biglybt3
biglybt2
biglybt

ferdnyc:
Wrap all calls to `setBackgroundMode(SWT.INHERIT_FORCE`) in an `if (Constants.isWindows) {}` conditional, to prevent them from overriding correct styling on other platforms.

Fixes #1002

### DIFF
--- a/uis/src/com/biglybt/ui/swt/TorrentMenuFancy.java
+++ b/uis/src/com/biglybt/ui/swt/TorrentMenuFancy.java
@@ -424,7 +424,10 @@ public class TorrentMenuFancy
 		shellLayout.marginWidth = shellLayout.marginHeight = SHELL_MARGIN;
 
 		shell.setLayout(shellLayout);
-		shell.setBackgroundMode(SWT.INHERIT_FORCE);
+		if (Constants.isWindows) {
+			// Windows SWT Bug: button and label BGs won't draw properly without INHERIT FORCE
+			shell.setBackgroundMode(SWT.INHERIT_FORCE);
+		}
 
 		topArea = new Composite(shell, SWT.DOUBLE_BUFFERED);
 		detailArea = new Composite(shell, SWT.DOUBLE_BUFFERED);
@@ -2512,7 +2515,11 @@ public class TorrentMenuFancy
 
 		HeaderInfo headerInfo = new HeaderInfo(id, runnable, composite);
 
-		composite.setBackgroundMode(SWT.INHERIT_FORCE);
+		if (Constants.isWindows) {
+			// Windows SWT Bug: button and label BGs won't draw properly without INHERIT FORCE
+			composite.setBackgroundMode(SWT.INHERIT_FORCE);
+		}
+
 		FillLayout fillLayout = new FillLayout();
 		fillLayout.marginWidth = 6;
 		fillLayout.marginHeight = 2;
--- a/uis/src/com/biglybt/ui/swt/devices/TranscodeChooser.java
+++ b/uis/src/com/biglybt/ui/swt/devices/TranscodeChooser.java
@@ -34,6 +34,7 @@ import com.biglybt.core.config.COConfigurationManager;
 import com.biglybt.core.devices.Device;
 import com.biglybt.core.devices.*;
 import com.biglybt.core.internat.MessageText;
+import com.biglybt.core.util.Constants;
 import com.biglybt.core.util.Debug;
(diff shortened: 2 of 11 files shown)


ferdnyc:
In my testing this appeared to have no ill effects, though I can't promise I found every involved code path.

Explicitly checked, and noticeably improved:
* `uis/src/com/biglybt/ui/swt/TorrentMenuFancy.java` – Active tab is now highlighted
* `uis/src/com/biglybt/ui/swt/views/MyTorrentsView.java` – Active Tag Button is now visually distinct
* `uis/src/com/biglybt/ui/swt/views/skin/StandardButtonsArea.java` – Dialog "OK" and "Cancel" buttons are styled natively

Explicitly checked, no difference observed:
* `uis/src/com/biglybt/ui/swt/views/columnsetup/TableColumnSetupWindow.java`
* `uis/src/com/biglybt/ui/swt/shells/opentorrent/OpenTorrentOptionsWindow.java`

TuxPaper:
I did some testing and can confirm that only Windows requires the INHERIT_FORCE.  OSX and the linux distros I tested on properly pull in the background of the parent without setBackgroundMode(..).

I'm going to sit on this pull for a day or two and ponder the best approach to fixing it.  The if statements all over the place is rather ugly.  Normally I'd create something in Utils, such as Utils.fixCompositeBugs(Composite), but doing that would still be ugly looking (a bit less ugly looking I suppose).  Creating a BiglyBTComposite class with the fix is out of the question, as there are Group and Shell.. plus it feels ugly too.

Perhaps a `Display.addFilter(SWT.Show, ...)` for Windows only, with code on the first visibility of the Composite.

TuxPaper:
for the record, I'm saying the setBackgroudMode code all over the place is already ugly, not that your pull request is ugly.  Well, I mean, it IS ugly, but ugly because the code is already ugly, if you catch my drift :)

TuxPaper:
Something like this seems to work. Oddly, the filter only picks up on things like Shell creation, and not every nested composite.  However, setting INHERIT_FORCE on the shell seems to make all nested widgets obey the rule.  I'm not sure if that's by design or some other code elsewhere.


```
Index: uis/src/com/biglybt/ui/swt/mainwindow/SWTThread.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- uis/src/com/biglybt/ui/swt/mainwindow/SWTThread.java	(revision 3fdaa2e807f457a7371c63b9caed00fb9f5eceed)
+++ uis/src/com/biglybt/ui/swt/mainwindow/SWTThread.java	(date 1556133537253)
@@ -238,6 +238,21 @@
 
 			MenuFactory.initSystemMenu();
 		}
+		
+		if (Constants.isWindows) {
+			display.addFilter(SWT.Show, event -> {
+				if (event.widget == null || event.widget.isDisposed()) {
+					return;
+				}
+				if (event.widget instanceof Composite) {
+					boolean needsFixupBG = event.widget.getData("DidFixupBG") == null;
+					if (needsFixupBG) {
+						((Composite) event.widget).setBackgroundMode(SWT.INHERIT_FORCE);
+						event.widget.setData("DidFixupBG", "");
+					}
+				}
+			});
+		}
 
 		if (app != null) {
 			app.runInSWTThread();
```

TuxPaper:
I've made my own changes similar to my previous comment and removed most of the setBackgroundMode(..).  Guess we'll see how well it works.

ferdnyc:
@TuxPaper I'll check out the HEAD and build it now, see how things look. If it solves the problem I can just close this PR instead of having to resolve all those conflicts! **#WIN**

ferdnyc:
Confirmed! Commit 696b0745 fixes #1002. Closing this PR.

PR #1029, opened 2019-04-24
Add Editorconfig files

ferdnyc:
This commit installs a top-level `.editorconfig` file which attempts to incorporate, to the extent supported, the rules present in the `PreferencesJavaCodeStyleFormatter.xml` config and/or laid out in the `CODING_GUIDELINES.md` file.

Most rules are applied to `*.java` files only, although a few are applied for all files:
```
[*]
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true
```

Markdown files are also explicitly configured to 2-space indents.

Additionally, local `.editorconfigs` are installed at each level of the tree which contains existing `.java` files in CRLF line ending format, to explicitly preserve that format on those files only.

For example, `core/src/com/biglybt/core/util/.editorconfig` contains:
```
[{AEMonitorOld.java,AESemaphoreOld.java,DataSourceResolver.java,protocol/URLConnectionExt.java}]
end_of_line = crlf
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,23 @@
+root = true
+
+[*]
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.java]
+tab_width = 2
+indent_style = tab
+indent_size = 2
+continuation_indent_size = 2
+end_of_line = lf
+spaces_around_operators = true
+spaces_around_brackets = outside
+curly_bracket_next_line = false
+indent_brace_style = K&R
+quote_type = double
+
+[*.md]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = false
--- a/core/src/com/biglybt/core/download/.editorconfig
+++ b/core/src/com/biglybt/core/download/.editorconfig
@@ -1,0 +1,2 @@
+[DownloadManagerOptionsHandler.java]
+end_of_line = crlf
--- a/core/src/com/biglybt/core/tag/.editorconfig
+++ b/core/src/com/biglybt/core/tag/.editorconfig
@@ -1,0 +1,2 @@
+[{TagGroup.java,TagGroupListener.java,TagUtils.java,TagWrapper.java}]
+end_of_line = crlf
--- a/core/src/com/biglybt/core/tracker/.editorconfig
+++ b/core/src/com/biglybt/core/tracker/.editorconfig
@@ -1,0 +1,2 @@
+[{AllTrackersManager.java,alltrackers/AllTrackersManager.java}]
+end_of_line = crlf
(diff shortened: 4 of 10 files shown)


lislei:
I've checked out this branch and tested it with my IntelliJ and I'm amazed on how it just magically picks those settings up.

What i tested is reformat of code - it seems to take the proposed configuration into account pretty well.
I had it picking up the utf8 encoding for java source once i typed in some special chars. 

I believe this is the better choice for solving line ending issues rather than using .gitattributes 

If I would suggest an improvement it would be to leave the .editorconfig files outside of the `src` folders - if not then they're pulled into the the produced jar file:
bilde

ferdnyc:
> I've checked out this branch and tested it with my IntelliJ and I'm amazed on how it just magically picks those settings up.

It's cool, right? And they're respected (to a lesser or greater degree) by nearly every editor out there, so no matter what someone's using on the code, they'll be covered by as much of the config as that editor supports. 

(Many of the entries, like `spaces_around_operators` and `quote_type`, are [extensions that IntelliJ and a few other editors have made](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#ideas-for-domain-specific-properties) to the grammar, they're not as [universally supported](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#current-universal-properties) as things like `end_of_line`, `indent_style`, or `tab_width`. But such extensions are fully blessed by the format in general, making it far more likely that support will broaden over time.)

I've _read_ that even GitHub's web-based tools will respect a limited set of `.editorconfig` properties, though I haven't tested that myself.

> If I would suggest an improvement it would be to leave the .editorconfig files outside of the `src` folders - if not then they're pulled into the the produced jar file:

Hah! That's... pretty dumb, actually. I never even thought to check that, good catch!

It's certainly an _option_ to move all of the configs up to the root of the project, at the risk of some very long paths in the file. I'm just wondering if there isn't a way we can just add an ignore to the build process that'll exclude them instead? I'll poke around, but I certainly have no objection to moving them if it comes to that.

ferdnyc:
Oh, I forgot to mention: One of the things I was doing, along side this, was looking at the history of the existing CRLF files. Roughly a _third_ of them have no history beyond their initial creation, so there's really no reason _not_ to just go ahead and fire up the steamroller: Their blame history is already flat.

Another quarter or so were _created_ within the past year or two, which TBH was the **main** impetus for this. It'll at least put at stop to any _further_ CRLF creep, regardless what gets done cleaning up the existing ones.

TuxPaper:
Surely there's a way to exclude `.editorconfig ` (and maybe any `.*` files) from the built JAR.  Afaik, parg has his own way to roll a JAR, so whether the file will be included or not in official JARs will be a mystery ;)

TuxPaper:
One annoying side affect is if I try to commit a file with a few lines of changes, I get a commit with a whole bunch of trailing whitespace removal.  I have to manually remove `trim_trailing_whitespace` from .editorconfig and undo the trims just to commit.

PR #1377, opened 2019-12-01
Update MessagesBundle.properties

nathanmuccino:
On 64-bit Windows 8.1 it appears that the 'newline' is not being rendered at all which results in the two words "tracker" and "without" to be smashed together into one word as 'trackerwithout'. I added a space to hopefully prevent this.


parg:
Thanks for the update - unfortunately if we make changes to a translation in the main message bundle then our CrowdIn translation automation goes ahead and deletes all existing translations for the key :( Looking at the code there is already a hack in there for this message key that is erroneously replacing the \n with "" - I've changed this to stick in a space which fixes the issue so your pull request isn't needed. Thanks again!

nathanmuccino:
Okay, that makes sense. Sound good, thank you, Sir!

nathanmuccino:
A better solution has been implemented by @parg

PR #1434, opened 2019-12-30
Drop support for GCJ in the startup script

ebourg:
The startup script for Linux can be simplified by dropping the logic for running with GCJ. GCJ is limited to Java 5 and no longer able to run BiglyBT, and it has been removed from GCC 7 and is no longer shipped with recent Linux distributions.


TuxPaper:
Thanks for fixing that up!  Glad to hear GCJ is virtually non-existent now (I suspected it was, but not enough to motivate me to fiddle with the start script, given my rusty bash skills)

Issue #1017, opened 2019-04-20
Failed to install DNS manager...


WishedGit:
Hello.

Since around B09 or B10 (not sure, not that long) I've been getting the following warning whenever I start BiglyBT:

"Failed to install DNS manager
This will impact functions such as forced DNS resolution prevention with SOCKS
Error: DNS SPI not loaded"

I've been through the settings and I can't remember changing anything. Doesn't appear to be impacting torrent functionality or performance though.

* Windows 10 Home 1809 17763.437
* Java 12 (64 bit)
* B1.9.0.1_B11/4 az3

parg:
That's related to https://bugs.openjdk.java.net/browse/JDK-8192780

WishedGit:
Thanks parg; didn't realise this was a Java error.

Downgraded (upgraded?) from Java 12 to latest Java 8 (u211 x64) and this fixed it.

parg:
Yup, we obviously want to move up from 8, and almost everything works fine with 12..., but we have limited resources and would rather Java brought the ability to intercept DNS calls!

Gatak:
Hi @parg Is there any downside to use Java 12, other than the DNS error? I use my own DNS server and don't use SOCKS.
Java 8 is rather old too, and afaik on my system, BiglyBT is the only that requires it.

Thanks :)

parg:
I don't know of any issues apart from that DNS one - it is more for trapping attempts to use the 'virtual' .i2p / .onion suffixes than SOCKS itself

Gatak:
OK great. Thank you. The Java bug is still open, I wonder if Oracle intends to fix it :/

sergiomb2:
AFAIK also we can't compile BiglyBT with Java 9 or 11[1]  . we have multiple examples on internet  ( error: package sun.net.spi.nameservice does not exist)  [2] . 

[1]
[ERROR] Failed to execute goal io.takari.maven.plugins:takari-lifecycle-plugin:2.0.0:compile (default-compile) on project biglybt-core: 16 error(s) encountered:
[ERROR] /builddir/build/BUILD/BiglyBT-2.7.0.2/core/src/com/biglybt/core/util/spi/AENameServiceDescriptor.java:[39:31] package sun.net.spi.nameservice does not exist
[...]

PR #2150, opened 2021-07-07
NameService SPI was removed in Java 9

sergiomb2:
and there is no replacement yet.
This commit fix compile of biglybt with Java 9 or Java 11

Also we have some references here: https://github.com/BiglySoftware/BiglyBT/issues/1017#issuecomment-873667879


parg:
We support Java 8 so this isn't going to be removed until then.

ferdnyc:
@parg You may not have to forever — [the latest Eclipse releases include an SWT that now requires Java 11+](https://help.eclipse.org/latest/index.jsp?topic=/org.eclipse.platform.doc.isv/samples/samples.html).

ferdnyc:
@sergiomb2 @parg

With #2611 merged, the exclusion of `AENameServiceDescriptor` when building under JDK 9+ is now completely automatic. Building BiglyBT doesn't force anyone to have to deal with it at all, anymore.

...I think this can probably be closed?

sergiomb2:
in Fedora we don't have takari-lifecycle-plugin , so I can't test the fix but LGTM 
I think we can close this PR

Issue #2157, opened 2021-07-12
[packaging question]  how I change the code to force instaltations in user dir ? 


sergiomb2:
biglybt when starts in a clean environment  try install azupdater and dorkbox-tray on /usr/share/java/biglybt when works well if install locally 

* Fedora 33 to Fedora 35 
* BiglyBT last version 
* Any Desktop

parg:
https://github.com/BiglySoftware/BiglyBT/wiki/Command-Line-Options has a section on options required to install a second instance of BiglyBT which shows what to set to force everything to a single dir

PR #2159, opened 2021-07-14
Let have a different INSTALL_DIR and PROGRAM_DIR

sergiomb2:
Regarding issue #2157 , I noticed that `-Djava.library.path=` and `-Dazureus.install.path` are always equal but can be different , let install plugins on homedir while main jar is installed java.library.path , fix my packaging problem .

### DIFF
--- a/core/src/com/biglybt/platform/unix/startupScript
+++ b/core/src/com/biglybt/platform/unix/startupScript
@@ -4,6 +4,7 @@
 AUTOUPDATE_SCRIPT=1 # change to 0 if you don't want your changes overwritten on next SCRIPT_VERSION change
 JAVA_PROGRAM_DIR=""	# use full path to java bin dir, ex. "/usr/java/j2sdk1.4.2/bin/"
 #PROGRAM_DIR="/home/username/apps/biglybt"	# use full path to BiglyBT bin dir
+#INSTALL_DIR="$HOME/.biglybt"
 JAVA_PROPS="-Djava.net.preferIPv4Stack=true"
 
 #######################################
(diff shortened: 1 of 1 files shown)


ferdnyc:
@sergiomb2 

My only suggestion (bikeshedders, assemble!) would be to go with something other than `INSTALL_DIR`, if the example path is going to be pointed at the user's `$HOME/.biglybt` directory. That's not really any kind of "install" dir. `CONFIG_DIR` or `SETTINGS_DIR` or `USER_PLUGINS_DIR` or something like that strikes me as more descriptive, if that's where it's expected to point.

And if that's _not_ where it's expected to point (like, say, it's pointing instead to a shared plugins path), then I'd still go with `PLUGINS_DIR` or something like that. `PROGRAM_DIR` and `INSTALL_DIR` sound like the exact same thing, to me.

sergiomb2:
> @sergiomb2
> 
> My only suggestion (bikeshedders, assemble!) would be to go with something other than `INSTALL_DIR`, if the example path is going to be pointed at the user's `$HOME/.biglybt` directory. That's not really any kind of "install" dir. `CONFIG_DIR` or `SETTINGS_DIR` or `USER_PLUGINS_DIR` or something like that strikes me as more descriptive, if that's where it's expected to point.
> 
> And if that's _not_ where it's expected to point (like, say, it's pointing instead to a shared plugins path), then I'd still go with `PLUGINS_DIR` or something like that. `PROGRAM_DIR` and `INSTALL_DIR` sound like the exact same thing, to me.

indeed, `USER_PLUGINS_DIR` is more intuitive , with USER_PLUGINS_DIR we may install plugins in our home directory and have biglybt installed on system, /usr/share/java/biglybt  . 

thanks for the suggestion

sergiomb2:
I changed the commit message to: [1] and the commit also already uses USER_PLUGINS_DIR as suggested.
   
   [1]
    We need a USER_PLUGINS_DIR    
    
    so that BiglyBT can be installed in the PROGRAM_DIR without write permission for the user, and the user can store their    
    own plugins in their USER_PLUGINS_DIR.

sergiomb2:
We need a USER_DIR to install BiglyBT globally on Linux

so that BiglyBT can be installed in the PROGRAM_DIR, where users do not have write permissions, while users can store their plugins and configuration in the USER_DIR.

PR #2343, opened 2021-11-27
Replace the statusbar LED GIFs with PNGs

ferdnyc:
I tried out the dark theme support recently, and discovered that while it mostly looks great, the "LED" GIFs used for the NAT and Ratio state indicators in the statusbar have incredibly poor transparency which makes them look like trash. They show massive amounts of speckling around the edges, against a dark background.

Here's how they look against white:

gif_light

And against black:

gif_dark

Ew, right? So, I drew new ones, as transparent PNGs, with specific focus on making sure the transparency of the pixels around the edge didn't bias the image towards any particular background color, and therefore look ugly against a _different_ background color.

My replacements, against white:

png_light

And against black:

png_dark

I don't _love_ them, and I think the gray and yellow could both be "brighter", but I figured I'd submit this first pass. I can make adjustments if anyone would prefer, just let me know.

This PR installs the new images into the `uis/src/com/biglybt/ui/icons/` directory (named `redled.png` for the `redled.gif` replacement, and so on...), and it changes the strings in `uis/src/com/biglybt/ui/icons/icons.properties` to point to the new icon files. Which I **think** is all that's necessary to make use of them, instead of the previous GIF icons? Please correct me if I'm wrong.

I didn't delete the `*led.gif` files. No real reason, I just didn't. Presumably they'll no longer get included in the build since they're not in the properties file, so they're only taking up space in the repo, not the JAR.


ferdnyc:
A question I'd been meaning to ask, if @Parg, @TuxPaper, or anyone else knows offhand — does SWT support SVG icons? Out-of-the-box / easily interchangeable for bitmapped ones?

If this were a Qt UI, I'd be inclined to just redo all of the icons as simple SVG files, instead. That way, the SVG engine can handle the transparency and the pixel antialiasing, and it solves the whole HiDPI issue for free. That's the route I took in [another project](https://github.com/OpenShot/openshot-qt/pull/4466), with good results. But that **was** a Qt UI, so I'm not sure if the same approach is workable in SWT.

parg:
Not sure about SVG, perhaps Tux knows, but I'll grab the pngs, thanks for the work!

PR #2420, opened 2022-01-19
Yet more dark-mode icon tweaks

ferdnyc:
This PR cleans up some more icons which are looking poorly in Dark Mode.

#### The globe icon for "Devices > Internet" (poor transparency)

The age-old story, speckly white pixels around the outer edges, when placed against a dark background. I re-shaded the border pixels, coloring them in with colors that better match the interior and using transparency, rather than brightness, to blend with the background.

Against white, and against black: globe_white globe_black
 
#### The "Archived Downloads" icon (semi-transparent black outline art)

This one, the existing icon consisted of nothing but an outline in black line art of varying degrees of transparency. It was nigh-invisible against the dark palette's default widget fill.

To flesh it out, I thresholded the outline to solid black, selected all of the interior pixels and hit them with a vaguely "wood-grain" pattern fill in shades of brown, and then rotated the hue of the original outline from black to a medium-dark brown, creating a vaguely wooden-looking inbox drawing from the outline.

Against white, and against black: archive_white archive_black
 
#### The Options icon

I saved this for last because I suspect it'll be the controversial one, if any. Based on the following observations:

1. This looks _especially_ like ass against the dark palette: image

1. Transparency issues aside, it's also _already_ a disaster of an icon, and TBH bears little to no actual resemblance to what it's supposed to represent (an isometric view of a blue gear). That we can see that _at all_ is the result of our brains doing some major interpolation based on a vaguely ellipsoid blue splotch.

1. Almost immediately once I started trying to fix the transparency/blending, I concluded the original icon (both sizes) is simply an unrecoverable mess. To blend well it would inevitably mean it looks even _less_ like what it already fails to represent, and it doesn't have the coherence to spare.

1. BiglyBT **already has** a perfectly fine icon representing a gear (which I've now made 100% background-agnostic; see first comment below). It's currently used for the Dashboard configuration icon: `images.sidebar.cog`. (This guy: cog_white )

Soooo... I just right up and **deleted** both `config.gif` and `config2.gif` from the `ui/images/sb/` directory with extreme prejudice, removed their mappings from the skin3 properties file, and replaced all of the code's references (a whole one each) to `images.sidebar.config` and `images.sidebar.config2` with `images.sidebar.cog`.

The same gray, non-isometric cog icon that sits to the right of "Dashboard" will now also be the icon to the left of "Options", and I personally think that's just fine.

...Thoughts?


ferdnyc:
After seeing how the cog icon looked against a white background above, I also filled its center hole with semi-transparent dark gray, so that it's more visible against a light background.

Updated `cog.png` against white, and black:  cog_white cog_black

adem4ik:
I like this PR. Just want to point a few more images, which look poorly in my Manjaro KDE:
- /uis/src/com/biglybt/ui/images/tb/tb_opentorrent-dark.png
image
- /uis/src/com/biglybt/ui/icons/openFolder16x12.gif
image

parg:
I'll get around to looking at this once I've fixed some other bugs.

FYI, you can have automatic light/dark mode icon switching by using properties entries with a ._dark suffix. for example

    dchat_gray=com/biglybt/ui/icons/statusbar/dchat_gray.png
    dchat_gray._dark=com/biglybt/ui/icons/statusbar/dchat_gray-dark.png

rather than having to come up with one icon that looks good in both themes

ferdnyc:
Ah, thanks @parg! I'll fix that.

ferdnyc:
@parg Hmmm. Would I then still change the code in `QuickLinksView.java` and `MainMDISetup.java` to use `images.sidebar.cog` directly, or would you prefer to leave the original (now aliased) properties? Does it matter?

ferdnyc:
> FYI, you can have automatic light/dark mode icon switching by using properties entries with a ._dark suffix.

Ooh, that's good to know in general.

The irony here (with the cog icon in particular) is that it _did_ look just fine in the dark palette. _Light_ backgrounds are where it kind of washed out, due to the transparent center hole. But, that should be fixed now.

There are definitely some more icons other than the blue gear that probably won't ever make a full transition to dual-palette life, though, so having the ability to sequester them on one side is helpful.

The folder icon @adem4ik pointed out (which I've also noticed, but have just kind of been avoiding TBH) is the first one that comes to mind. It's probably fine to leave it alone for the light skin, but it'll need a separate dark-palette version. Even if it's just a simple photonegative of the current icon. (Ooh, that'd mean mostly-orange folder icons... which could look kind of OK!)

ferdnyc:
Updated to alias the old properties instead of removing them, as @parg advised. I left the code changes in for now, but I can also drop those if needed.

ferdnyc:
So, once I mentioned the photonegative-orange version of the blue folder icon, I had to try it.

It looks... sort of awesome, but sort of weird at the same time. I don't know how I feel about it.

So then I did an orange colorization of the blue folder icon, _without_ inverting the colors, and cleaned up the transparency around the edges to make it look better against a dark background. Which looks... _very_ different from the negative version. I can't decide if I like either, or which one I prefer! So I'll appeal to the wisdom of the comons: You decide!

Here they both are, against a larger than usual black field because I find it helps...

Photonegative:

negative


Recolored/reshaded:

recolor

(I guess if nothing else I'd have to lighten up the photonegative version a bit. But it does have a kind of awesome **#METAL!** 🤘 look to it.)

ferdnyc:
@adem4ik 

> * /uis/src/com/biglybt/ui/images/tb/tb_opentorrent-dark.png
>   image

The irony there, as you can see from the filename, is that that one's _already_ a dark-palette version. Honestly I think that one looks bad for the same reason  the blue gear does: It's a terrible, ugly icon in any color scheme. **#KILLITWITHFIRE**

parg:
BTW, so nice to have you guys contributing to things, yay :)

ferdnyc:
Hmm, I never heard any responses on the folder icon(s), so (other than converting the existing one from GIF to PNG) that never got addressed.

Eh, maybe I'll just keep it blue. Having done the transparency blending on the reshaded orange version, it shouldn't actually be any harder to fix the original blue version. It's not nearly as hopeless as I feared / as the gear icon was.

And I'm tempted to just set it as the "Open Torrrents" toolbar button's icon as well, because that current icon is the pits.

ferdnyc:
Whoops. Looks like I broke the find-files dialog's folder icon, with my conversion to PNG. I missed this property mapping:

https://github.com/BiglySoftware/BiglyBT/blob/eed062bd599863f93ea1fb62ce2d353454379a4b/uis/src/com/biglybt/ui/skin/skin3_dlg_findfiles.properties#L52

I'm about to submit a PR fixing its transparency anyway, I'll take care of that as well.

PR #2426, opened 2022-01-22
Removing reflection use in NetUtils

lislei:
Removed this forward compatibility code as the Android SDK level 15 / JDK8 API is the current baseline. 
Wrapped it in a separate method. 

[NetworkInterface.getByIndex(int)](https://docs.oracle.com/javase/8/docs/api/java/net/NetworkInterface.html#getByIndex-int-) was introduced with Java 7.

### DIFF
--- a/core/src/com/biglybt/core/util/NetUtils.java
+++ b/core/src/com/biglybt/core/util/NetUtils.java
@@ -20,8 +20,6 @@
 
 package com.biglybt.core.util;
 
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
 import java.net.*;
 import java.util.*;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


core/src/com/biglybt/core/util/NetUtils.java
```diff
@@ -549,37 +547,15 @@ at whatevercalled getNetworkInterfaces()
...
-			if (list.size() > 0) {
-				return Collections.enumeration(list);
-			}
-		} catch (NoSuchMethodException ignore) {
+		final List<NetworkInterface> list = new ArrayList<>();
+
+		collectNetworkInterfacesByIndex(list);
+
+		if (!list.isEmpty()) {
+			return Collections.enumeration(list);
```
lislei, line 555: This "Java 7 preview" block of code made the runtime return at this line when running Java 7 or higher.

I believe the next block was for JDK 6 and earlier only, making that a tautology on the current baseline. 
Should I go ahead and remove it ?

ferdnyc: I mean, it only returned here **if** the list was non-empty. Which doesn't seem like it's _guaranteed_ even still, so I'd say it's far from tautological.

**Most** of the time the code will never get this far, period. It'll return the valid list produced by `getNetworkInterfaces()` on line 526 (524 in your PR), none of the code past that will even execute, and all will be right with the world.

But on some system where `getNetworkInterfaces()` _fails_, can we say for sure that `getByIndex()` won't fail as well? Especially when the code is just _guessing_ indices.

It's clear from the docs that the `getByIndex()` method was **meant** to be used by code that _already_ had a known valid index, most likely retrieved by previously calling `.getIndex()` on an interface object and storing it somewhere. And the docs explicitly refuse to guarantee anything about the index values themselves, other than that they're integers >= 0. So there's nothing stopping some implementation from using high and/or randomized index values that the `getByIndex()` approach won't ever find.

In which case, guessing names _could_ theoretically still work as an absolute last-resort fallback.

ferdnyc: > So there's nothing stopping some implementation from using high and/or randomized index values that the `getByIndex()` approach won't ever find.

Oh, nope, wait, I'm wrong about that, sorry. Since the code will (in theory) try every integer if `.getByIndex()` is available, the only way that would fail to eventually find **something** is if `.getByIndex()` was implemented but couldn't actually be used to look up interfaces, which they're not allowed to do.

In that case, I don't see any reason the name-guessing code would still be needed.

lislei: Thank you for assessing this question. Your initial comment makes good sense. 

I now believe this (and the next) error path are "best efforts" of reacting to runtimes throwing you exceptions from the underlying OS. They both give no guarantees of finding _any_ interfaces what so ever. In the worst case - the fall trough - is simply re-throwing the initial exception. 

I'll add a brief documentation describing the two error paths.

PR #2427, opened 2022-01-22
Dark-mode folder icon fixes

ferdnyc:
This PR:

1. Fixes the folder icon in the find-files dialog, which I inadvertently broke in #2420 .
2. Fixes the transparency of that icon to be background-agnostic
3. Replaces the weird folder icon used for the "Open Torrents" toolbar button with the one mentioned above, the same one that's used everywhere else a folder icon is shown.

@parg: 

One thing I'm not clear on, how critical are dimensions for things like button icons? The "Open Torrents" toolbar button's previous icon was 21x15 with 2px of whitespace on both the left and right; the replacement is a tightly-packed 16x12. So it's 1px narrower and 3px shorter in terms of actual image content, plus it lacks 4px of extra width padding.

Do I have to do anything to ensure that it ends up aligned correctly in the button, or will it get positioned automatically? _Can_ I even do anything to align it the same way the previous icon was aligned, or do I need to create a separate 21x15 version of the PNG?

I don't think it'll really matter if the button is 5px narrower or the text is 2px closer to the icon, but if it's not going to be vertically centered or if it'll affect the height of the button itself then that's probably not ideal.


ferdnyc:
Note to myself: The following context-menu screenshot contains both a smaller version of the Archive "inbox" icon I missed when I re-did the larger one, and more fugly icon-transparency issues than you can shake a GIMP pattern at:

image

ferdnyc:
(**Including** a refugee blue gear! **#KILLITWITHFIRE** **#BuryItInTheWoodsWithAShovelTHENBURYTHESHOVEL**)

parg:
I can't say in general how important the various icon dimensions are unfortunately - I'd need to go check all the alignments and whether or not the overall layout is explicitly sized or relies on the icon's size. For example, the open-torrents icon top left might have its overall height set by the icon, by the text or explicitly so it matches the search box height on its right.

Safest approach is to keep the size the same and pad with transparency...

ferdnyc:
> Safest approach is to keep the size the same and pad with transparency...

*nod* My intuition settled on the same thing, I made a 21x15 version of the icon last night.

I think I also need to make a padded version of the cog icon, because the one shown in the Quick Links button for the preferences is a little off center now. Not glaringly, but (especially because it's so subtle) it'll bug the crap out of ME.

It's like looking at keming, you can't _not_ see it!

ferdnyc:
Progress update:
image

Plus I fixed the transparency on Edit Trackers & did dark-mode variants of the various column/row mode icons (none of which are visible here); the import icon got a dark-mode variant as well (in addition to export).

I think that's enough GIMP for one evening.

ferdnyc:
Fixed add_tracker (because I'd already done edit_tracker, which it used as a base), but... less like a space alien deposited some sort of mutant growth into one of its points: add_tracker

ferdnyc:
I am not looking forward to wrestling with that spigot icon (Set up/down speed.) I opened it at 3200% zoom (my usual icon-editing size) and all I could think was, "I can see the sailboat!"

image

parg:
Feel free to turn it into something completely different!

PR #2439, opened 2022-01-29
Last of the dark mode icon updates

ferdnyc:
OK, I think this closes out the set of dark-mode icon fixes I'll be submitting. There was one more I _wanted_ to do, but ultimately decided to pass on: the media server icon. It just does not lend itself to hand-correction. Plus, it's "not so bad" as it is, compared to some of the others..

I also didn't fix the magnet icon (from the library context menu). That's because _I couldn't **find** it!!_ I have to assume it's contained in the magnet plugin itself, because it sure isn't among the icons in this repo. Anyway, that icon is also "not that bad" in its current form.

The spigot icon (for Set up/down speed), I replaced with an all-new icon:

image

It's meant to evoke a speedometer. (Though I realize the color progression is more typical of a tachometer.) I'm not sure it really looks like _either_, but... `¯\_(ツ)_/¯`

And I did a _bunch_ of dark-mode variations or device icons. (Including some that probably are never used anymore — Wii, anyone? Blackberry? — but, meh.) ...Of course, that's assuming `._dark` variants _work_ for the device sidebar icons, something I wasn't able to test. But, worst case is they don't get used and nothing changes, NBD.

The PSP icon, I didn't do a dark _variant_, but I reshaped a lot of the icon's features so it should be a bit more evocative of a PSP on both light and dark backgrounds. (Though, again, that's probably a pretty useless icon nowadays.)


ferdnyc:
> I have to assume it's contained in the magnet plugin itself, because it sure isn't among the icons in this repo.

Ah, I was right, it's here: 
`core/src/com/biglybt/plugin/magnet/icons/magnet.gif` image

So, technically it _is_ in this repo.

parg:
Really done? :) Good job finding the magnet and thanks for all the work!

ferdnyc:
@parg
> Really done? :)

Yeahhhh... well, _maaaaaybe_ really-really done. I thought I was, and in the sense that I'd finished everything I was planning to do, I am.


Buuuut, I'd never used the Meta Search interface before, and when I felt adventurous and _did_...

image

Wow, those all look like hot ass.

So, I may at some point find motivation to do something about  that. (Particularly the audio icon, which hurts to look at.)

But for now, I'm done. Content-type icon fixes are a separate project I may or may not start at some point in the future. (Whatever I have to tell myself...)

ferdnyc:
Ooh, speaking of! I just experimentally clicked "Download" on one of my results, and was presented with this dialog:

image

...I'll open a new issue for that.

PR #2443, opened 2022-01-30
Optimize PNG images losslessly with FileOptimizer

adem4ik:
PNG images were optimized losslessly with [FileOptimizer](https://nikkhokkho.sourceforge.io/static.php?page=FileOptimizer). It saved about 300 Kbytes:
image

P.S. You may need to review the changes using Firefox, my Vivaldi wasn't able to show them correctly.


parg:
Hmm, I ran them all through "pngcrush" relatively recently, I'm surprised that they can further be optimized that much!

PR #2590, opened 2022-06-13
Blocked IP log initial scroll positioning

ferdnyc:
In separate commits, this PR:

1. Adds code to automatically position the block log to the _bottom_ of its scroll, when the "Blocked IPs" window first opens. (The log is time-ordered, with the most recent entries at the bottom, so it makes sense to automatically make those items visible in the initial view.)

1. Trims a  whole bunch of trailing whitespace from the lines in that file. (My editor actually did that automatically, but I figured I'd submit it since I was opening a PR anyway. Like I said, it's a separate commit.)

I haven't actually tested this, myself — building a BiglyBT jar locally just requires way too much effort to test a one-line change. (Especially since I've upgraded to a newer Fedora release since the _last_ time I did that, and distro upgrades pretty reliably break my ability to successfully build the project, forcing me to reconstruct a build environment from the ground up every time.)

It _should_ work, since it's identical to code @parg used in the SFEDF window, and we know it works there. But, yaknow... life rarely gives guarantees, eh? `¯\_(ツ)_/¯`

### DIFF
--- a/uis/src/com/biglybt/ui/swt/BlockedIpsWindow.java
+++ b/uis/src/com/biglybt/ui/swt/BlockedIpsWindow.java
@@ -130,16 +130,16 @@ BlockedIpsWindow
 
     Label	lblUnban	= new Label(window,SWT.NULL );
     Messages.setLanguageText(lblUnban,"ConfigView.section.ipfilter.unban");
-    
+
     Text	textUnban	= new Text(window,SWT.BORDER );
-    
+
     Button btnUnban = new Button(window,SWT.PUSH);
     Messages.setLanguageText(btnUnban,"Button.unban");
-    
+
     btnUnban.addListener(SWT.Selection, (ev)->{
     	IpFilter filter = core.getIpFilterManager().getIPFilter();
     	BadIps badIPs = core.getIpFilterManager().getBadIps();
-    	
+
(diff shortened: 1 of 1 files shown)


ferdnyc:
> I haven't actually tested this, myself — building a BiglyBT jar locally just requires way too much effort to test a one-line change.

I keep making threatening noises (...at...myself?) about CI tooling, and then nothing ever comes of it. I'd definitely fire me, if I were getting paid for this.

ferdnyc:
@parg
> Looks like it might work ;)

Hrm. If the change made it into `_B36`, then it doesn't work. :-(

ferdnyc:
Hmmm. Perhaps the `.setSelection()` has to come _after_ the `window.open()` (or at least the `window.layout()`) here, to be effective:

https://github.com/BiglySoftware/BiglyBT/blob/5bc8c380e357876ade4770ba8ad3e659840c7320/uis/src/com/biglybt/ui/swt/BlockedIpsWindow.java#L275-L278

parg:
Ah, you needed

textBlocked.setSelection(textBlocked.getText().length());

fixed in next beta :)

ferdnyc:
D'oh! I missed that `txtInfo` in the `TextViewerWindow` is a `Text`, not a `StyledText`. :disguised_face: 

> fixed in next beta :)

\o/

PR #2611, opened 2022-06-30
Maven: Auto-exclude AENameServiceDescriptor based on JDK version

ferdnyc:
Building under JDK 1.8.0 / JDK 8, the `com.biglybt.core` classes can be compiled with the DNS SPI code in `...core.util.spi.AENameServiceDescriptor`. Thanks to the shim classes in `AENameServiceJava9.java` and `...Java12.java`, the compiled code can even be run under later JRE releases.

When building under JDK 9 or higher, `AENameServiceDescriptor.java` is uncompilable; the only option, [as @parg has noted](https://github.com/BiglySoftware/BiglyBT/issues/1017#issuecomment-873861491), is to "Just exclude the file from the build".

There's no reason that has to be a manual process, thanks to Maven's profiles feature. This PR adds a new profile, activated only when the Java version Maven is being run under is 9 or higher, that automatically excludes the `AENameServiceDescriptor.java` file from the build.

My tests were run on my Fedora 36 system, with multiple JRE releases installed.

With JRE 1.8.0, Maven compiled 2652 sources in core/:
```console
$ JAVA_HOME=/usr/lib/jvm/jre-1.8.0-openjdk mvn -B clean
$ JAVA_HOME=/usr/lib/jvm/jre-1.8.0-openjdk mvn -B package
--------8<--------
[INFO] --- takari-lifecycle-plugin:2.0.0:compile (default-compile) @ biglybt-core ---
[INFO] Previous incremental build state does not exist, performing full build
[INFO] Compiling 2652 sources to /home/ferd/Projects/BiglyBT/core/target/classes
--------8<--------
[INFO] Compiled 2652 out of 2652 sources (17450 ms)
```

With JRE 11, Maven compiled 265**1** sources successfully:
```console
$ JAVA_HOME=/usr/lib/jvm/jre-11-openjdk mvn -B clean
$ JAVA_HOME=/usr/lib/jvm/jre-11-openjdk mvn -B package
--------8<--------
[INFO] --- takari-lifecycle-plugin:2.0.0:compile (default-compile) @ biglybt-core ---
[INFO] Previous incremental build state does not exist, performing full build
[INFO] Compiling 2651 sources to /home/ferd/Projects/BiglyBT/core/target/classes
--------8<--------
[INFO] Compiled 2651 out of 2651 sources (39938 ms)
```

(JRE 17 and 18 builds fail for other reasons, it seems the code is not yet fully compatible with those Java releases.)

Note that no other configuration was required on my part, just changing which JRE Maven executed under. Assuming this works the same for Parg and doesn't interfere with the _official_ build process in any way, it should make life easier for   distro packagers and others who want to build BiglyBT themselves.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -34,8 +34,9 @@
 		</resources>
 		<plugins>
 			<plugin>
-				<groupId>org.apache.maven.plugins</groupId>
-				<artifactId>maven-compiler-plugin</artifactId>
+				<groupId>io.takari.maven.plugins</groupId>
+				<artifactId>takari-lifecycle-plugin</artifactId>
+				<extensions>true</extensions>
 				<configuration>
 					<source>${java.version}</source>
 					<target>${java.version}</target>
@@ -43,5 +44,28 @@
 			</plugin>
 		</plugins>
 	</build>
+
+	<profiles>
+		<profile>
+			<id>java-9</id>
+			<activation>
+				<activeByDefault>false</activeByDefault>
+				<jdk>[9,)</jdk>
+			</activation>
+			<build>
+				<plugins>
+					<plugin>
+						<groupId>io.takari.maven.plugins</groupId>
+						<artifactId>takari-lifecycle-plugin</artifactId>
+						<configuration>
+							<excludes>
+								<exclude>**/core/util/spi/AENameServiceDescriptor.java</exclude>
+							</excludes>
+						</configuration>
+					</plugin>
+				</plugins>
+			</build>
+		</profile>
+	</profiles>
 
 </project>


ferdnyc:
I'd _like_ to auto-exclude the shim classes as well (`AENameServiceJava9.java` and `AENameServiceJava12.java`), since AIUI they're pointless if `AENameServiceDescriptor` isn't present at all. (Someone please correct me if that's wrong!) However, they're harder to exclude as they're imported by the `COConfigurationManager`.

More-clever suggestions welcome, but my best idea so far for how to handle that would be to move all of the `com/biglybt/core/util/spi/*` files out of `core/src/` and into another directory, say `core/nameservice_jre8`. Then, in `core/nameservice_dummy/com/biglybt/core/util/spi/`, we could place bare-bones, no-op shims for `AENameServiceJava{9,12}.java`, both containing just an `init()` method that immediately returns `true`.

Instead of a profile that auto-_ex_cludes one file in JRE 9+ builds, two profiles would _in_-clude one or the other directory, depending on the JRE level detected.  That way, there would always be code to satisfy the import lines in `COConfigurationManager`:

https://github.com/BiglySoftware/BiglyBT/blob/60fd14a71099a76338582531634f6a98040ec9d9/core/src/com/biglybt/core/config/COConfigurationManager.java#L34-L35

...So nothing would break there, and the unnecessary hackery the shims perform wouldn't be bothered with if the JAR wasn't built with `AENameServiceDescriptor` in the first place.

parg:
Hmm, I'd rather just use reflection to invoke the AENameServiceJava9/12::init() methods, I'll do that

ferdnyc:
> I'll do that

Cool, because other than knowing that it's a big buzzword in Java, I know precious little about reflection. :grin: 

Feel free to commit to this branch, I just sent you a collab invite.

parg:
Too late, already committed to master/origin :)

sergiomb2:
for me,  yes , I will try to test it soon 
Thank you

ferdnyc:
Hmm, unfortunately this isn't _quiiite_ there yet.

The **build** process is fine, and the resulting JAR works nicely, but even with #2616 applied BiglyBT still pops up an alert message on startup, complaining that it's unable to install the DNS manager. Which is exactly what I hoped would be avoided, if I just _removed_ the manager completely. (In addition to making it possible to compile the code at all.) There's nothing to install! 

Now we just need to get the code to realize that. Seems there are still a couple more hooks left to (conditionally) disconnect.

I also need to take another look at the build failures under JDK 17, to get an idea of what it would take to make the code compatible with that release too.

ferdnyc:
Oh! Actually, it looks like I accidentally fixed JDK 17 builds in #2616! That wasn't even on my radar, but I'll take it!

parg:
BiglyBT attempts to install the DNS SPI code or hack around it for a reason, so I'm not sure that just because the code isn't there anymore it shouldn't still complain that something isn't quite right...

PR #2616, opened 2022-07-02
Exclude NS shims, service file in Java 11+ builds

ferdnyc:
Now that the `AENameServiceJava{9,12}.java` sources can be excluded from the build, this PR expands the automated profile logic to do so.

It also ensures that the `META-INF/services/sun.net.spi.nameservice.NameServiceDescriptor` file will only be included in the JAR when it's built under JDK 8 or lower. (When the build includes `com.biglybt.core.util.spi.NameServiceDescriptor`, in other words.)


parg:
Just noticed that this has feked things up - moving services out of META-INF has borked things for Java8 users :(

PR #2634, opened 2022-07-11
Fix JDK17

ferdnyc:
This PR  uses the copy/rename Maven plugin to copy two translation files to their updated names, when building under JDK17+. This allows the tests to pass.

Now that JDK17 builds are successful, run them on all three platforms in CI, same as other versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,18 +11,11 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.continue-on-error }}
 
     strategy:
-      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-10.15, windows-latest]
-        java-version: [ '8', '11' ]
-        continue-on-error: [false]
-        include:
-          - os: ubuntu-latest
-            java-version: '17'
-            continue-on-error: true
+        java-version: [ '8', '11', '17' ]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


ferdnyc:
bullwinkle-this_time_for_sure

ferdnyc:
Hmm. While everything is now passing, to make Windows and macOS more happy it seems like it would be a really good idea to sort out automatic inclusion of the `libOSXAccess_10.5.jnilib` / `aereg64.dll` native files into the final packaging.

ferdnyc:
> This PR uses the copy/rename Maven plugin to copy two translation files to their updated names, when building under JDK17+. This allows the tests to pass.

Hmmm. Now I'm wondering whether we should _always_ do this, for the packaging. The question is, if the  code is _compiled_ under JDK 8/11, but then _run_ under JDK 17, which names will it expect those translation files to have? If it's the legacy names, that's fine. If it's the _new_ names, then this should be an always-step, not a JDK17+-only step.

I guess what we really need is a Hebrew- or Indonesian-language user who can run the app under JDK 17 and let us know whether the interface is properly localized.

ferdnyc:
@parg: This _would_, at a slight cost of extra space taken up in the JAR, since the two renamed translation files will be present under _both_ names in the final JAR. (It literally just copies the file to the alternate name, leaving the old naming in place.)

There are code changes that can be made under JDK17 to have to continue using the legacy language codes; I sort of hate to do that because it feels backwards, but it's probably the only workable solution that doesn't involve duplicating the data. (The link I posted to the other PR about the language-code changes has info on how to continue using the legacy codes in JDK 17+.)

ferdnyc:
(Actually, I should check the artifact sizes — depending how smart zip compression is, two files with different names but identical contents might not even increase the archive size by more than a few bytes for the extra dictionary entry.)

Issue #2587, opened 2022-06-10
change color of progress column


johnvolfovits:
* win 7
* 3.0.0.0
![Untitled1](https://user-images.githubusercontent.com/107244103/173020596-b6322ce8-07c3-4825-afb7-e829fd9613c7.png)

* For *nix: Desktop Environment
how to change color of progress column
the color of this column used to be blue and when torrent complete it get green. But now the color doesn't change and by this color the note that shown on column such as percentage of complete, speed and etc are completely unreadable. i want to change color myself or at least change the color to what it was before

parg:
Can't explain that one... my Windows 7 is showing blue/green for downloading/seeding download, grey for those in other states.

There is a transparent image that is drawn over the background to give it the vague 3D look, I wonder if something's weird about that.

parg:
The next beta (3001_B35) will have a right-click column header option to disable the 3D look, see if that helps.

johnvolfovits:
is that has a relate to win theme?
is there any way to set that myself

parg:
your guess is as good as mine. B35 will be available in 30 mins

johnvolfovits:
its not working
is there any way to set that myself

parg:
what's not working?

parg:
You installed B35. Then you right-clicked on the "Progress" column header and deselected the "3D Look" option?

ferdnyc:
I don't think this is the 3D look — that wouldn't explain why the _text_ (which is drawn **on top** of the shaded bar) is also gray.

Hey @johnvolfovits, you don't have the preference
Interface > Color Scheme > Color Overrides > Force non-system colors to grayscale
turned on, by any chance, do you?

(I'm not sure that would affect this either, as those colors are practically _hardcoded_ — but I also don't know how that option actually works, so maybe it would.)

johnvolfovits:
hi 
yes that's right i turned i on and when i unchecked this it get fix
thanks anyway

PR #2647, opened 2022-07-17
WIP: HELP WANTED: Progress column redesign

ferdnyc:
I'm submitting this not because I'm hoping to get it merged (please DO NOT MERGE!! in its current state), but to ask for help with the code in a forum conducive to others looking at and offering comments on the code in its current state.

Inspired in large part by #2587, I've been working on a total redesign of the "Progress" column bar graph and associated information. For the most part, it's going well, and in fact it's almost ready, but I'm stuck on one last problem that I can't seem to figure out.

The new design:

1. Does away with the 3D-look "overlay" PNG. Instead, it draws the 3D shading using `fillGradientRectangle()` with highlight/lowlight colors produced by `getLighterColor()` (which, despite its name, can also produce darker shades).
2. Adds several additional color customizations to the skin file, replacing some hardcoded colors. It also adds dark-mode variants of the existing colors, so that the column matches the dark palette when activated. (MOSTLY!)
3. Because everything is now drawn programmatically and there are no more fixed sizes, the bar graph will dynamically resize itself to the row height. Not in the way it did before (by centering itself vertically), but actually drawing taller or shorter depending on the available space.
4. The font size will also grow when given more vertical space, although there's a constraint in place based on column aspect ratio that prevents it from getting too large to fit the necessary text. (Which does mean that, at larger row heights, sizing the column wider will allow the font to grow even larger.)

Now, the problem I'm having: Everything looks fine when there's only enough height for one-line formatting, but when the height increases enough the column keeps the current behavior of moving the status text to a second line below the graph. And that's where my troubles come in.

For some reason, it appears I'm getting passed a graphics context with a foreground color of _green_ and a background color of _white_, whether the table is in dark mode or light mode. I can't figure out how I'm supposed to match the foreground and background colors of the table row, to draw the second-line text. Here's how things look, in light & dark modes:

Screenshot from 2022-07-17 18-09-11
Screenshot from 2022-07-17 18-09-04
Screenshot from 2022-07-17 18-08-56
Screenshot from 2022-07-17 18-08-49

Screenshot from 2022-07-17 18-10-46
Screenshot from 2022-07-17 18-10-56
Screenshot from 2022-07-17 18-11-01
Screenshot from 2022-07-17 18-11-07

Can anyone spot my probably-stupid mistake? @parg? @TuxPaper?


ferdnyc:
Oh, yeah, for the record: In-progress transfer shading:

Screenshot from 2022-07-17 18-43-49
Screenshot from 2022-07-17 18-42-51

I might need to lighten the blue bar, and tone down the 3D shading a bit, for light mode / smaller heights.

parg:
It's almost as if it were useful... 

BTW, are these CI builds filling up github's disks with bigabytes of data each commit or are old builds purged at some point?

ferdnyc:
> BTW, are these CI builds filling up github's disks with bigabytes of data each commit or are old builds purged at some point?

Yes and yes. The artifacts take up an obscene amount of space (over a gig per build, currently), which GitHub is apparently perfectly happy to provide. They're auto-purged after 90 days, currently, though you can lower that number in the repo Settings... somewhere.

ferdnyc:
I've always assumed that _someday_ they'll end up instituting some sort of quota system, when the disk space consumed by CI grows to outstrip the US national debt. But they haven't made any noise about that so far, amazingly, and when it happens it won't be our fault. There are projects that publish _hundreds_ of gigs per build, it's insane.

ferdnyc:
> It's also weird that, in your 2-line version, the "STOPPED" is taller than the bar itself.

OK, that part I can reproduce, on Linux. I must've overbalanced somewhere.

> Very narrow rows don't render too well though

That part I can't, though. It must be a Windows font-sizing issue. I'll have to add in some rounding for the font sizes.

The weird-colors thing is also still happening to me, it must be a Linux SWT thing I guess. I also noticed that when the row is _selected_, the colors are fine:

image

parg:
Here's by kubuntu VM

image

ferdnyc:
Motherf----! WTF?  Arrrgh.

Well, I think I at least figured out the font-sizing issue. I set the minimum font size to 16px, hoping to combat tinyfont issues like what you're seeing in the bar on Ubuntu. But that doesn't work (that progress-bar font is _definitely_ smaller than 16px), because it looks like Linux SWT will automatically scale the font down to fit inside the paint area if it's too large. So, effectively, the font size you specify doesn't really matter.

But Windows is using the actual 16px font it was told to use (crazy, right?), which means it's being forced way too large for the available space in some instances.

parg:
write-once-run-everywhere as they said

parg:
mahahahah!

ferdnyc:
@parg 

OK, I've greatly simplified the sizing logic, doing away with the aspect-ratio constraints and fixing the size of the second line, now it's either visible or not. The text in the bar is still sized based on the bar height, but only up to a fixed maximum of 24px, because nobody needs to see the completion percentage in 48px text no matter how tall/wide their progress cells are. The 3D-effect shading also shrinks down slightly for shorter row heights, to be less overpowering in tight quarters.

Only remaining issue is that I'm seeing "java.lang.Error: SWT Resource was not properly disposed" messages for the font allocations, despite _trying_ to be careful about always calling `.dispose()` on the previous allocation(s) whenever I'm updating the font size.

Part of me thinks that it would be better to compute the font sizes and allocate them in a resize listener (is there such a thing, in SWT?), rather than doing them inside the paint routine that gets called so frequently. I have a feeling the sheer number of calls to the paint routine, all potentially updating `static` class variables containing the font info, might be trampling each other. (Then again, a resize listener would also be called many times when updating the cell dimensions, so maybe it wouldn't make much difference. Plus, all of the paint calls should be happening on the SWT thread, so they shouldn't be concurrent anyway, right?)

Could be that I'm just doing something wrong with how I manage the font allocations, based on my (lack of) understanding of how SWT works. Like, maybe something's causing the font to be duplicated somewhere, and I'm not cleaning it up correctly. (I removed the code that set `fontSecondLine` equal to `fontText` in one-line mode, since I thought that might be causing trouble. Now I just use `gc.setFont(showSecondLine ? fontSecondLine : fontText)` to set the font before rendering the status text, but maybe that's worse somehow since the call to `gc.setFont(fontText)` will be redundant in one-line mode.)

parg:
Maybe don't bother disposing the fonts in remove()? Not sure it is anything to do with your dispose error but you don't need to tidy them up there and I'm not sure if that method is invoked on the SWT thread.

All significant SWT stuff occurs on the single SWT thread so you don't need to worry about concurrent access to static vars etc.

When are you getting that error message? On closedown?

ferdnyc:
@parg
> Maybe don't bother disposing the fonts in remove()? Not sure it is anything to do with your dispose error but you don't need to tidy them up there and I'm not sure if that method is invoked on the SWT thread.

Mmm, I added that to try and mitigate the messages, it definitely isn't the _cause_. But I'll take it out, because it didn't help any.

> When are you getting that error message? On closedown?

Nope, it's happening multiple times for line 531 (the `getFontWithHeight() call for `fontText`), with every change in row height. If the second line is showing, I get them for line 564 (the `getFontWithHeight()` call for `fontSecondLine`) as well.

ferdnyc:
UGGGH. F'in GitHub.
> This is a scheduled macOS-10.15 brownout. The macOS-10.15 environment is deprecated and will be removed on August 30th, 2022. For more details, see https://github.com/actions/virtual-environments/issues/5583

Guess it's time to upgrade the CI environment to macOS 11. Shouldn't make much difference, I don't think.

parg:
oh, it is a bug :)

getFontWithHeight doesn't dispose of the font generated by FontUtils.getFontPercentOf(baseFont, pct); when "runAgain" is true...

parg:
Try with https://github.com/BiglySoftware/BiglyBT/commit/af9e23f7867bf17b524d7a47521d68d8409dbb87

parg:
Ready for mergin?

ferdnyc:
@parg
I'll give it a whirl now. But if that's fixed, then yeah, the dispose errors were the only outstanding issue from my POV.

ferdnyc:
Yup! Everything looks A-OK now, if you feel my code looks up to snuff. (I'm the last person to judge that, beyond the fact that it compiles cleanly and runs without any glaring errors. But, subtler bugs? My nickname ain't exactly "Raid".)

parg:
I think you need to ensure that when the row-height is set to 2 or more then the widget goes into two-line mode:

image

Otherwise the "simple" Library list view isn't what it oughta be

parg:
You could also change the default for the new column to "not 3D"?

vadzimk:
BiglyBt used to achieve a minimalist and at the same time full of features design but this new "redesign"  is just HURTING MY EYES
Exactly how you saw it — with capital letters shouting at me directly from BiglyBt.
Moreover it is hard to match with your eye which row that particular DOWNLOADING refers to: to the below or to the above.
image

Please make a it one-level cell as it used to be before.

PR #2652, opened 2022-07-21
CI: Switch to macos-latest

ferdnyc:
GitHub has started browning out `macos-10.15`, as it's going away at the end of August.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-10.15, windows-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


ferdnyc:
Hmm. That works fine, but I want to update the script to eliminate the separate Build step. I just noticed that, for some reason, it's _recompiling_ all of the sources during the Packaging step, so it's just a waste of time to do a separate Build step. (I figured takari's incremental build manager would let it skip compilation the second time through, but that's not working for whatever reason.)

ferdnyc:
Well, one Windows job and one Mac job got bogged down, so overall the run was 3 minutes _slower_ than the previous one. But the Linux jobs were each ~ 20 seconds faster (impressive considering they were all under 1m30s _before_), and I suspect on average we'll see at least some speedup from eliminating the redundant step.

PR #2656, opened 2022-07-23
ProgressETA: Lower margin height to 1

ferdnyc:
On my headless machine (running LxQt, instead of GNOME Shell), I noticed that the new progress bar looks a little "squished". The code sets the cell margin to 3, which seems excessive. (I think the old code's drawing routines may have ignored the margins.)

The current code, without these changes:
image


ferdnyc:
Much better:
image

(This is a narrower column, unfortunately I resized it between the two screenshots. But the text is definitely larger.)

ferdnyc:
Added a couple of height tweaks, because the progress display wasn't actually going to 2-line mode until the row was 2.5 lines tall (again, in LxQt).

parg:
I'm changing the default for the 3d stuff to false for the new column

parg:
Hmm, on Windows it still isn't going to two-line mode when the row height is set to two lines...

PR #2786, opened 2022-11-08
ProgressETA: Avoid NPEs with FakeTableCell

ferdnyc:
I noticed storms of these in my debug log earlier:

```text
[04:39:21] [stderr]   java.lang.NullPointerException
        at com.biglybt.ui.swt.columns.torrent.ColumnProgressETA.cellPaint(ColumnProgressETA.java:486)
        at com.biglybt.ui.swt.views.table.impl.FakeTableCell.invokeSWTPaintListeners(FakeTableCell.java:370)
        at com.biglybt.ui.swt.views.table.impl.FakeTableCell.doPaint(FakeTableCell.java:1092)
        at com.biglybt.ui.swt.views.table.impl.FakeTableCell.paintControl(FakeTableCell.java:1166)
        at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:234)
        at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
        at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5810)
        at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1529)
        at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1555)
        at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1538)
        at org.eclipse.swt.widgets.Control.gtk_draw(Control.java:3891)
        at org.eclipse.swt.widgets.Scrollable.gtk_draw(Scrollable.java:365)
        at org.eclipse.swt.widgets.Composite.gtk_draw(Composite.java:496)
        at org.eclipse.swt.widgets.Widget.windowProc(Widget.java:2482)
        at org.eclipse.swt.widgets.Control.windowProc(Control.java:6834)
        at org.eclipse.swt.widgets.Display.windowProc(Display.java:6118)
        at org.eclipse.swt.internal.gtk3.GTK3.gtk_main_do_event(Native Method)
        at org.eclipse.swt.widgets.Display.eventProc(Display.java:1552)
        at org.eclipse.swt.internal.gtk3.GTK3.gtk_main_iteration_do(Native Method)
        at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4469)
        at com.biglybt.ui.swt.mainwindow.SWTThread.<init>(SWTThread.java:438)
```

`FakeTableCell` constructs a stub `TableRow` in response to `.getTableRow()`, and that object's `.getView()` returns `null`. So, it can't be dereffed unilaterally.

`FakeTableCell` also returns `-1` in response to `.getMaxLines()`, tho, so sequestering the `.getLineHeight()` lookup behind a test for `.getMaxLines() > 1` avoids the `lineHeight` lookup via `getView().getLineHeight()`, and thus the NPE.

### DIFF
--- a/uis/src/com/biglybt/ui/swt/columns/torrent/ColumnProgressETA.java
+++ b/uis/src/com/biglybt/ui/swt/columns/torrent/ColumnProgressETA.java
@@ -483,7 +483,6 @@ public class ColumnProgressETA
 			return;
 		}
 
-		int lineHeight = cell.getTableRow().getView().getLineHeight();
 		final boolean showSecondLine;
 		int alignSecondLine;
 		Color fgSecondLine;
@@ -493,6 +492,14 @@ public class ColumnProgressETA
 
 		boundsProgressBar.width--;	// adjust as includes pixel for column line :(
 		
+		final int lineCount = cell.getMaxLines();
+		int lineHeight;
+		if ( lineCount > 1 ) {
+			lineHeight = cell.getTableRow().getView().getLineHeight();
+		} else {
+			lineHeight = cellBounds.height;
+		}
+
 		if (cellBounds.height >= Math.min( minTwoLineHeight, lineHeight*2 )) {
 			showSecondLine = true;
 			boundsProgressBar.height -= secondLineHeight;


uis/src/com/biglybt/ui/swt/columns/torrent/ColumnProgressETA.java
```diff
@@ -493,6 +492,14 @@ public void cellPaint(GC gc, TableCellSWT cell) {
 
 		boundsProgressBar.width--;	// adjust as includes pixel for column line :(
 		
+		final int lineCount = cell.getMaxLines();
+		int lineHeight;
+		if ( lineCount > 1 ) {
+			lineHeight = cell.getTableRow().getView().getLineHeight();
+		} else {
+			lineHeight = cellBounds.height;
```
ferdnyc, line 500: This won't interfere with ProgressETA activating two-line mode when it can, even in single-text-line height rows. It initially seems like it would, and it took me a bit to convince myself it wouldn't.

It **seems** like a mistake that the code won't do the `lineHeight` lookup from the view unless `.getMaxLines()` is at least 2, but that's actually correct. It doesn't need the real line height in those cases. In fact, that's the whole point of this `Math.Min()` call: ` Math.min( minTwoLineHeight, lineHeight*2 )` .

Unless we can fit 2+ lines of text in the row, we want the `showSecondLine` decision to be based on `minTwoLineHeight`. Using a `lineHeight` equal to the cell height ensures that it will be. Whether that's because the row only fits one actually-rendered line of text at the given font size, or because it's a FakeTableCell with no real height at all, doesn't matter to the code here.

PR #2796, opened 2022-11-17
ProgressETA: Simplify 2-line checks, remove ColumnProgressETAOld.java

ferdnyc:
The more I thought about it, the more I realized my https://github.com/BiglySoftware/BiglyBT/pull/2786#discussion_r1016682779 was correct — we don't need the cell line height at all, we just want to know if it's at least 2 lines tall. That info's available _without_ having to look up the view, via the table cell's `.getMaxLines()`.

So, this PR simplifies that check.

It also removes `ColumnProgressETAOld.java`, as I think we're pretty all-in on the rewrite. (The old code is still available in the git history for `ColumnProgressETA.java` anyway.)


uis/src/com/biglybt/ui/swt/columns/utils/TableColumnCreatorV3.java
```diff
@@ -18,9 +18,6 @@
 
 package com.biglybt.ui.swt.columns.utils;
 
-import java.lang.reflect.Constructor;
-import java.util.*;
-
```
ferdnyc, line 23: Sorry, IntellijIDEA likes to optimize imports.

uis/src/com/biglybt/ui/swt/columns/utils/TableColumnCreatorV3.java
```diff
@@ -39,6 +36,12 @@
 import com.biglybt.ui.swt.views.table.utils.TableColumnCreator;
 import com.biglybt.ui.swt.views.tableitems.mytorrents.*;
 
+import java.lang.reflect.Constructor;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
```
ferdnyc, line 44: (It replaced them with this.)

PR #2872, opened 2023-02-13
Stringify tracker announce logging

ferdnyc:
The announce logger was dumping a raw ByteArrayOutputStream into the log; it should be converted to string first.

Before:

Screenshot from 2023-02-12 23-08-59


ferdnyc:
@parg: Ping? Pretty trivial change, but it'll make the logs look so much less messy.

parg:
I've been away...

ferdnyc:
@parg That's OK, I realized this is actually pointless, it will change nothing.

The problem, really, seems to be that some tracker announce messages aren't printable (they contain binary data — I don't know _what_, perhaps encryption keys?). So, maybe they shouldn't be logged at all (not the contents, anyway). But they'll come out just as garbled this way as they do now.

Another option would be to run the messages through a `charset` instance that just drops all non-`us-ascii` characters, instead of printing garbage. The log messages will _look_ better, but they won't make any more sense, and if anything they'll be even less useful. (Not that they're necessarily useful now.)

ferdnyc:
I'll leave this open, but convert it to a draft until we figure out what (or if) should be done.

ferdnyc:
Huuuh. I... can't do that?

Are Draft PRs disabled in the repo?

parg:
Might be better just to remove the logging of the raw response

ferdnyc:
@parg That's what I was thinking, too. Even in the best cases where the whole message _is_ printable, it's a serialization like:

> `d8:completei8e8:intervali1800e12:min intervali1800e5:peers0:`

Which... I can certainly imagine some rare instances where that data _**might**_ be useful to have. (And, it **is** a _debug_ log message, after all.) But even then, it probably makes more sense to log the parsed values from whatever class processes the message, not just stuff the _raw_ message bytes into the log.

I'll update the PR to drop — or, comment out? — that logging call. (But I'll leave the one in the exception handler up above it alone; some noise in a log is a fair tradeoff for content dumps when an error occurs.)

ferdnyc:
> And, it **is** a _debug_ log message, after all.

Actually, that's wrong, isn't it? It's logged at Info level. Well, then it _definitely_ goes, there's nothing informational about that message.

parg:
Feel free to hack it out :)

ferdnyc:
@parg Done, it's no more.

PR #2895, opened 2023-02-27
IpFilter: Tweak logging

ferdnyc:
This PR consists of two commits which make minor/housekeeping changes to the logging in the IpFilter implementation:

1. Split up IPFilter LogIDs
    - All filter-list management will still be logged to `LogIDs.CORE`. `CORE` is still the default `IpFilterImpl.LOGID`.
    - Individual IP block activity will be logged to `LogIDs.NWMAN`, instead.
2. Lower block success logs to INFO

It never made sense to me, that the logs for blocked IPs were in the "core" logging stream, while everything else to do with connection setup and teardown was logged as "nwman". So, the first commit moves _only_ the logs for blocked-IP additions to "nwman", while leaving the logging for management of filter lists in the "core" stream.

It also never made sense to me that successful blocks (a perfectly normal course-of-operations event, when a filter list is defined) were being logged at the "WARNING" level. There's nothing abnormal or actionable about a successfully-blocked IP, so for the log message to be calling attention to itself is just distracting.

### DIFF
--- a/core/src/com/biglybt/core/ipfilter/impl/IpFilterImpl.java
+++ b/core/src/com/biglybt/core/ipfilter/impl/IpFilterImpl.java
@@ -491,7 +491,7 @@ IpFilterImpl
 	      if ( addBlockedIP( new BlockedIpImpl( ipAddress, match, torrent_name, loggable), torrent_hash, loggable )){
 
 		      if (Logger.isEnabled())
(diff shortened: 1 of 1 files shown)


ferdnyc:
Hahaha! And the CI for this is [Actions job 666](https://github.com/BiglySoftware/BiglyBT/actions/runs/4286213316) on the repo. A bad omen? :grinning:

PR #2935, opened 2023-04-10
Recreate aereg with minimal changes in Visual Studio 2022

### LINKED ISSUE #2922: Working on aereg (libWIN32Access) Using Visual Studio
While working on #1870 I needed to examine the native code in libWIN32Access. Apparently it is a Visual Studio solution with only one project. The project seems to have been created with a really old version of Visual Studio (likely Visual Studio 6 as evidenced by .dsp and .dsw files) and later updated to Visual Studio 2010 (as evidenced by the contents of .vcsproj file). When importing it in Visual Studio 2022 I had to upgrade the Platform Toolset and Windows SDK version used to build the project and even then there were other compilation errors that needed to be dealt with. I also examined the Git log and saw no updates to this directory ever since the initial commit.

Is there any other method of working with this part of the code base or if, as the Git log suggests, no one is actually working with it, are there any plans or guidelines for bringing it to such a state?

soroosh-chabi:
This is to fix #2922.

Created a new project with Visual Studio 2022 and imported source files from this project into it and made some changes to get it to build. The old link libraries and header file that originally belonged to Windows SDK were no longer needed because the new Windows SDKs already have them.

Both x86 and x64 artifacts are generated as `aereg.dll` but in different directories and since no single BiglyBT installation needs both of these files, the code was changed so it wouldn't append `64` for x64 architectures.


parg:
Excellent, thanks for the work!

PR #973, opened 2019-03-30
poc - tool for crowdin translation updates

lislei:
@TuxPaper 
I [previously I mentioned](https://github.com/BiglySoftware/BiglyBT/commit/c81885c7e55fe8917efb91ef184731b3e0e456eb#r32921680)
 a possible way of updating changes to the translation files with Crowdin.

I wrote a ugly, nasty, poc that collects and assembles a zip file for uploading to crowdin.
ref I hope you would consider trying it out to see if it's a possible way of automating manual changes to the language files. 

Run with
 `gradlew stageLocalizations -Pcommit=ddaad11a`

Then upload the zip with crowdin.

PR #2999, opened 2023-07-02
Add crowdin translation-upload (feedback) action

ferdnyc:
This PR grew out of discussion in #973, regarding feeding translations back to Crowdin when they're updated locally. The code here is an attempt to make that work. It is untested, because it **cannot** be tested until some prerequisites are met.

This PR will not work and **SHOULD NOT BE MERGED** unless the repository is first configured with the necessary Crowdin credentials so that the action has access to upload into the Crowdin project.

Obtaining that access requires someone with admin access to both the Crowdin and Github sides of the project to do the following:

1. Go to https://crowdin.com/project/biglybt/settings#api and obtain the crowdin Project ID (API v2)
2. Go to https://crowdin.com/settings#api-key and generate a personal token to upload to the project
3. Go to https://github.com/BiglySoftware/BiglyBT/settings/secrets/actions and create two secrets:
    * `CROWDIN_PROJECT_ID` should store the project ID
    * `CROWDIN_PERSONAL_TOKEN` should store the generated token

Only _after_ that is done, will it be possible to **TEST** this PR. Because it currently contains the setting `dryrun_action: true`, it still won't do anything that would affect the Crowdin project.

If things look sane, then we can remove that line and test it for real.

cc: @parg @TuxPaper


ferdnyc:
Oh, right: Because Github sensibly won't run newly-created actions from PRs, this will have to be merged before any testing can be done. So, I guess, once the secrets are created, it would probably make sense to merge this with the `dryrun_action: true` line still in it. That'll let us see if things look right, at which point a subsequent commit/PR can remove the dryrun line.

ferdnyc:
I added `workflow_dispatch` to the action as a trigger method, so that it will be possible for project admins to manually run the workflow from the Actions tab of the repo. That'll probably be useful at least for testing, because it's configured to only _automatically_ run when a commit changes files in the `core/src/com/biglybt/internat/` directory. (So, very infrequently.)

ferdnyc:
That failure looks transient, it died in the wiki tests on Windows for some reason.

ferdnyc:
I'll force-push the branch so the CI restarts, see what that failure was all about.

PR #3014, opened 2023-07-21
Fix various typos in specific subdirectories

luzpaz:
Fixes user-facing and non-user-facing typos including documentation typos. 

The PR has been split in to several commits for reviewing convenience.


luzpaz:
JFYI, if there hesitation about merging because it would bloat the code blame, consider adding a `.git-blame-ignore-revs` file: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

luzpaz:
There are more typos but distributed throughout a several dozen different directories. Is there an interest for me to add a catch-all commit that includes these misc. typos ?

parg:
Thanks! Don't get too carried away, there must be a bazillion typos and I don't have the bandwidth to check all the fixes!

ferdnyc:
What does it say, that this actually doesn't seem as bad as I expected? :rofl:

parg:
lol

PR #3029, opened 2023-08-19
Shorten default smoothing window to 20s (min 10s)

ferdnyc:
This PR (in three separate commits)

1. Changes the configuration for smoothing window, by:
    - Reducing the default `SMOOTHING_UPDATE_WINDOW` to `20`, down from `60`
    - Reducing the _minimum_ value to `10`, down from `30`
    - (Leaving the maximum value at `30 * 60`)

2. Removes an unused import from `core/src/com/biglybt/core/global/GlobalManagerStats.java`

3. Updates the build version in the pom files to `3.4.0.1-SNAPSHOT`

#### Motivation

The current window values/constraints were arrived at many years ago, when speeds were much lower and swarms were much smaller. With modern broadband speeds and data rates, even a multi-gigabyte torrent might _complete_ downloading in under 60s, for many users.

When entire torrents download in a matter of seconds or single-digit minutes, because of the current 60s window the initial "ramp-up" time (when fewer peers are connected and data rates are suppressed) continues to bias the ETAs for far too much of the transfer, throwing off the completion estimates until the download is nearly complete.

Using a shorter window allows the ETA estimates to approach reality much more quickly, once the torrent "hits its stride" at maximum transfer rate.

#### Impact

**Note:** I'm not clear on which, but either _all_ existing users, or only existing users who've already customized the setting in **Options > Statistics > General > Smoothed average window size [secs]**, will be unaffected by this change unless they use that setting to reduce their window size, after installing a build of BiglyBT that includes these changes.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -8,7 +8,7 @@
 	<parent>
 		<groupId>com.biglybt</groupId>
 		<artifactId>biglybt-parent</artifactId>
-		<version>3.1.0.1-SNAPSHOT</version>
+		<version>3.4.0.1-SNAPSHOT</version>
 	</parent>
 
 	<dependencies>
--- a/core/src/com/biglybt/core/global/GlobalManagerStats.java
+++ b/core/src/com/biglybt/core/global/GlobalManagerStats.java
@@ -61,12 +61,12 @@ GlobalManagerStats
 
 	/**
 	 * Smoothed Send Rate, including data and protocol, based on
-	 * "Stats Smoothing Secs" (default to 60s, min 30s)
+	 * "Stats Smoothing Secs" (default to 20s, min 10s)
 	 */
 	public long	getSmoothedSendRate();
 	/**
 	 * Smoothed Receive Rate, including data and protocol, based on
-	 * "Stats Smoothing Secs" (default to 60s, min 30s)
+	 * "Stats Smoothing Secs" (default to 20s, min 10s)
 	 */
 	public long	getSmoothedReceiveRate();
 
--- a/core/src/com/biglybt/core/global/GlobalManagerStats.java
+++ b/core/src/com/biglybt/core/global/GlobalManagerStats.java
@@ -24,7 +24,6 @@ import java.net.InetAddress;
 import java.util.*;
 
 import com.biglybt.core.download.DownloadManager;
-import com.biglybt.core.util.CopyOnWriteList;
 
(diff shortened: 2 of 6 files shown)


ferdnyc:
(Hopefully I've done this right, and not missed any of the locations where these values need to be adjusted.)

parg:
Guess we'll give it a go and see what happens :)

parg:
Hmm, actually the default is 120 seconds - in ConfigurationDefaults.java

    def.put( "Stats Smoothing Secs", new Long( 2*60 ));

and your change won't affect this...

I'll go ahead and fix this but am left wondering if such a big drop from 120 secs to 20 is reasonable? What do you think?

ferdnyc:
@parg 

I still think it makes sense... like I said, these days a torrent can _finish_ in ~60 seconds. (Certainly, **well** under 120!)

Before, what I'd see when starting a hundreds- or thousands-of-seeds, ~1GB torrent was:
1. During the initial ramp-up, with only 1 or 2 slow peers connected, the rates would be in the 100 – 200 KBps range, and the ETA would grow to several hours.
2. Over the next 5-10 seconds, more, faster seeds would connect (up to my configured max of 20/torrent), and the download rates would quickly rise to multiple MBps. 
3. Much more slowly, the ETA would start working its way back down.
4. For the rest of the first minute, as completion approached 80%, 90%, the ETA would drop from hours to tens of minutes, then single-digit minutes
5. Only when the torrent made it to > 98% complete and was mere seconds away from completion, would the ETA (very briefly) correctly reflect that, only for a moment or two just before reaching 100%.

With the window shortened, once the transfer rate gets up into multiple MBps, the ETA will drop to small numbers of minutes... which is not too far off from how long it will _actually_ take.

When it comes to much larger, or especially poorly-seeded, torrents, there will probably be a _bit_ more fluctuation in the ETA as speeds shift up and down. But I suspect even then, it'll just be a matter of the estimates being more honest about how current transfer rates impact the predicted completion time.

parg:
It is used for a few things, I've played about a bit and have increased the default to 30 seconds. Base averages are currently over a 10 second interval, 30 seconds at least gives a level of smoothing to that...

parg:
BTW, rightly or wrongly, the same smoothed rates are used in the choke/unchoke algorithm and request allocation by the piece picker, they're not just for stats display.

PR #3347, opened 2024-08-08
Add headless profile for compilation w/o GUI

sot-tech:
Useful for platforms without (or orphaned support) SWT like solaris/illumos.

Process start will produce `java.lang.UnsatisfiedLinkError: Could not load SWT library` (not critical)
because of adding SWT libraries to `-cp` in shell script, but i don't know easy way to check if they need to be added.


parg:
<exclude>**/biglybt/plugin/**</exclude> matches more than you want, I'll move the SWT specific plugin components to "com.biglybt.ui.swt.plugin" which is where they should be

parg:
Ah, OK, if the exlusion only applies to the "ui" part then fine. I've moved the classes anyway, should still work

sot-tech:
deleted the exclusion to avoid a double interpretation

Issue #3497, opened 2025-03-10
to bash or not to bash


fkobi:
Hello,

with the recent inclusion of the distribution files in `assets/linux/` I have noticed that some of the scripts were using `bash` when they could be using just the posix `sh`.

Here's the current state of the shebangs for sh:
```console
$ rg "^#\!.*sh$"
assets/linux/unregisterBiglyBT
1:#!/usr/bin/env bash

assets/linux/updateBiglyBT
1:#!/usr/bin/env bash

assets/linux/registerBiglyBT
1:#!/usr/bin/env bash

mvnw
1:#!/bin/sh

core/src/com/biglybt/platform/unix/startupScript
1:#!/usr/bin/env bash
```
for `{un,}registerBiglyBT` it's trivial but `updateBiglyBT` & `startupScript` require more work.

One thing is clear: they are not written **for** bash -- they are shell scripts with occasional bash syntax which makes them not portable.
Right now we do not the portability (and speed, for people with alternative `/bin/sh`) of the posix shell nor the speed and the convenience of bash.

I think it should be decided if the project should:
- go all in for `bash`
- use bash only on Linux (FreeBSD does not ship it by default)
- stick to the posix `sh`

I am willing to handle all this myself but I do not want my work to go to waste so please decide on the shell policy for this project.

also as a sidenote: Does it make sense to have the _unix_ startup script written for bash?
as far as I know it's the standard only on GNU/Linux.

TuxPaper:
If I had to make an educated guess, these scripts started off simple posix shell scripts, but as time and went on, different devs went with whatever was easier or what they knew (bash syntax).

I don't have an issue with having them as posix sh, but I'll have to discuss with with parg since he may know something I don't.

ferdnyc:
I've been looking at this, and at the two PRs opened (one merged, one in draft). My take is, it doesn't seem to make sense to me that _one_ script be a `bash` script, while the _others_ are straight Bourne shell scripts.

In fact, that mismatch seems like it would be the worst of both worlds. It increases the likelihood that some user on a system where `bash` isn't a given (Ubuntu's bundled bare-bones is apparently `dash`, these days) will now succeed in _installing_ BiglyBT (because the register/unregister scripts are `/bin/sh`-compatible), but then  updates will fail due to a lack of `bash` to run `updateBiglyBT`. Previously, when all the scripts used `bash`, at least a lack of it would surface earlier.

[...]

PR #3498, opened 2025-03-10
Use `sh` instead of `bash` when it's easy

fkobi:
The normal POSIX sh is simpler and can be faster than bash.
For more details please see https://github.com/BiglySoftware/BiglyBT/issues/3497.


TuxPaper:
This is nice. Thanks for your work!

PR #3597, opened 2025-09-11
Move issue template to supported location

ferdnyc:
The issue template has stopped working, which seems to be caused by it not being in the officially-supported location. I guess they dropped support for legacy paths.

This should fix it.


ferdnyc:
Damn, still not working. I think it may need another config file to go along with the template. I'll fix it up shortly. I have templates enabled in like a dozen repos — If even I can figure it out, it can't be that hard!

parg:
thanks :)

PR #3636, opened 2025-11-23
New version of unbundled Apache Commons using Apache Commons Text

sergiomb2:
New version of unbundled Apache Commons using Apache Commons Text
    
The replacement is mostly straightforward except for a small case.
Apache Commons Text uses apache-commons-lang3 which are actively maintained,
providing better stability and future compatibility.


parg:
If it works don't break it?

sergiomb2:
Fixed after added org.apache.commons:commons-text dependency to pom.xml and core/pom.xml

sergiomb2:
Use apache-commons-lang3 and apache-commons-text

PR #3639, opened 2025-11-23
Ensure JAVA_VERSION is set when JAVA_PROGRAM_DIR is defined

sergiomb2:
Always call look_for_java to set JAVA_VERSION    
    
This ensure that `--add-opens java.base/java.net=ALL-UNNAMED` is added to JAVA_ARGS for Java 11 or higher [1]    
    
Previously, when JAVA_PROGRAM_DIR was set, look_for_java was not called,    
and therefore JAVA_VERSION was not set.    
    
 [1]

     if echo "$JAVA_VERSION" | grep -E "^(1[1-9]|[2-9][0-9])\." > /dev/null ; then    
       JAVA_ARGS="${JAVA_ARGS} --add-opens java.base/java.net=ALL-UNNAMED"    
     fi

### DIFF
--- a/core/src/com/biglybt/platform/unix/startupScript
+++ b/core/src/com/biglybt/platform/unix/startupScript
@@ -175,6 +175,8 @@ if [ "$JAVA_PROGRAM_DIR" == "" ]; then
 	if ! look_for_java ; then
 		exit 1
 	fi
+else
+	check_version
 fi
 
 # Java 11+ needs --add-opens, which borks some older java versions


sergiomb2:
Hi,
The issue here is that when we set JAVA_PROGRAM_DIR as I do on Fedora, the script runs look_for_java only when "$JAVA_PROGRAM_DIR" is empty [1]. If it doesn’t run look_for_java, then JAVA_VERSION is not set, and when JAVA_VERSION is not set, the script fails to add --add-opens java.base/java.net=ALL-UNNAMED to JAVA_ARGS, which is very important for Java 11+ versions 

[1] 
```
    # locate and test the java executable
    if [ "$JAVA_PROGRAM_DIR" == "" ]; then
        if ! look_for_java ; then
            exit 1
        fi
    fi
    
    # Java 11+ needs --add-opens, which borks some older java versions
    if echo "$JAVA_VERSION" | grep -E "^(1[1-9]|[2-9][0-9])\." > /dev/null ; then
      JAVA_ARGS="${JAVA_ARGS} --add-opens java.base/java.net=ALL-UNNAMED"
    fi
```

parg:
But "look_for_java" can update "JAVA_PROGRAM_DIR", surely you don't want this? Possibly you want to call "check_version"?

sergiomb2:
Correct , just call check_version when  JAVA_PROGRAM_DIR is defined also works

PR #3657, opened 2025-12-30
biglybt.desktop without placeholders

sergiomb2:
Essentially equivalent to running
sed -i -e '/^#!/d' -e 's|${installer:dir.main}/||' -e 's|\.svg||' -e 's/Application;//g' biglybt.desktop,
plus adding Keywords=file transfer;P2P;

### DIFF
--- a/assets/linux/biglybt.desktop
+++ b/assets/linux/biglybt.desktop
@@ -1,12 +1,12 @@
-#!/usr/bin/env xdg-open
 [Desktop Entry]
 Name=BiglyBT
 GenericName=BitTorrent Client
 GenericName[ru]=Клиент сети БитТоррент
-Exec=${installer:dir.main}/biglybt %U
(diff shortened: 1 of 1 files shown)


TuxPaper:
The insertion variables are replaced in the installation process.  If you open up the .desktop file installed, it should have the full real path instead.  Feels safer to keep it this way in case people move the .desktop file to another folder.

sergiomb2:
> The insertion variables are replaced in the installation process. If you open up the .desktop file installed, it should have the full real path instead. Feels safer to keep it this way in case people move the .desktop file to another folder.

But how is the replacement done, and where is it performed?

sergiomb2:
have we any correct way to replace the placeholders ? I just notice that usually executable and icon aren't in the same directory and the placeholders are always installer:dir:main 
Exec=${installer:dir.main}/biglybt %U
Icon=${installer:dir.main}/biglybt

PR #3734, opened 2026-04-18
Fix search filter bleed-through between Tags Overview and Torrent Details

### LINKED ISSUE #3733: Search boxes in Files views have weird interactions between Tags Overview and Torrent Details
This is surely my fault, at least indirectly, for the request I made in #3696.

Now that the table layout for Tags Overview's Files view and the Torrent Details Files view are separated, there's a weird interaction with the search boxes in those two views. As far as I can tell, here's how it now works:

1. Filling in the search box at **Tags Overview > Files** (after selecting a tag) limits the rows displayed to match the term entered, as normal.

2. That search term becomes the _default_ for new **Torrent Details > Files** views opened after that point.

3. Editing the search term in a **Torrent Details > Files** will update the list in that individual view to show the matching rows, but it _won't_ change the default for new **Torrent Details > Files** views. They'll still come up with the last search that was entered at **Tags Overview > Files**.

4. Even more confusing, the search **ISN'T** persisted at **Tags Overview > Files** itself — it will come up with an empty search box each time it's closed and reopened, even if it was closed with search terms filled in. (They'll be persisted to **Torrent Details > Files**, instead.)

5. Because of the previous item, closing and opening **Tags Overview > Files** will reset the default search for new **Torrent Details > Files** views, since the previous stored value will be cleared out when **Tags Overview > Files** opens with a blank search box.

Things get even more confusing when **Tags Overview** and **Torrent Details** are open at the same time:

* Editing the search at **Tags Overview > Files** has no effect on any open **Torrent Details > Files** views or their search boxes.
* Editing the search at **Torrent Details > Files** will _blank_ the search box in  **Tags Overview > Files** if it's currently open, even if new search terms were entered at **Torrent Details > Files**.

### System block
```
Java 21.0.10 (64 bit)
  Red Hat, Inc.
/usr/lib/jvm/java-21-openjdk

SWT v4968r13, gtk/3.24.52, zoom=100, dpi=96
Linux v6.19.10-200.fc43.x86_64, amd64 (64 bit)
B4.0.0.1_B29/4 az3 en
```

ThroneEmpire:
Hopefully fixes https://github.com/BiglySoftware/BiglyBT/issues/3733

### DIFF
--- a/uis/src/com/biglybt/ui/swt/views/FilesView.java
+++ b/uis/src/com/biglybt/ui/swt/views/FilesView.java
@@ -154,6 +154,12 @@ public class FilesView
   public static final String MSGID_PREFIX = "FilesView";
 
   private String view_sub_id;
+
+  private String
+  getViewKey()
+  {
+  	return( view_sub_id != null ? MSGID_PREFIX + "." + view_sub_id : MSGID_PREFIX );
+  }
   
   private DownloadManager[] managers = new DownloadManager[0];
 
@@ -529,7 +535,7 @@ public class FilesView
 				
 				if ( map != null ){
 					
-					Map<String,Object> view = map.get( "FilesView" );
+					Map<String,Object> view = map.get( getViewKey() );
 					
 					if ( view != null ){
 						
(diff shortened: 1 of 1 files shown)


parg:
Thanks, looks reasonable

